PR #5, opened 2018-08-28
Add config for travis

williamjamir:


.travis.yml
```diff
@@ -1,25 +1,20 @@
...
 python:
-  - 3.6
-
+- 3.6
 install:
-  - pip install -U pip
-  - pip install -U tox-travis
-
-# Command to run tests, e.g. python setup.py test
+- pip install -U pip tox-travis
```
nicoddemus: You should update `pip` first, otherwise there's no point in doing it at all 😉 

```
- pip install -U pip
- pip install tox-travis
```

nicoddemus: (Just applied the change myself)

PR #7, opened 2018-09-20
Add data_types and data_model

williamjamir:


alfasim_sdk/data_model.py
```diff
@@ -0,0 +1,78 @@
+import functools
+from typing import Optional
+
+import attr
+
+from alfasim_sdk.data_types import BaseField
+
+
+def container_model(*, MODEL: type, CAPTION: str, ICON: Optional[str]):
```
nicoddemus: please mind the code conventions, parameters should all be lower case.

alfasim_sdk/data_types.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@attr.s
+class Enum(BaseField):
+    value: List[str] = attrib()
+    initial: str = attrib(validator=optional(instance_of(str)), default=None)
+
+    @value.validator
+    def check(self, attr, values):
+        if not isinstance(values, (list, set)):
```
nicoddemus: Not sure we should allow `set`s, as they don't keep order.

williamjamir: ✔️ `set` removed

PR #9, opened 2018-10-04
Add enable_expr option

williamjamir:
This pull request adds the `enable_expr` option on the `BaseField`.

Although this property is being introduced on the `BaseField`, the `enable_expr` it's an _optional_ property that the user can define or not.

In order to make this property be _optional_ and at the same time be _inherited_ by all objects, was necessary to include `kw_only=True` argument on the `attrs `declaration.


nicoddemus:
Good job!

PR #10, opened 2018-10-11
Keeping the declaration order.

williamjamir:
This pull request changes the usage of `dir(class_)` to  ` class_.__dict__` to keep the declaration order intact

Also, this pull request adds a check to do not allow underscore at beginning of attributes names.


nicoddemus:
Good work!

PR #13, opened 2018-11-13
Update Hooks initialize and finalize and add new functions to API

ggrbill:


alfasim_sdk/hook_specs.py
```diff
@@ -1,14 +1,14 @@
 from hookman.hooks import HookSpecs
 
 
-def initialize(ctx: 'void*', v1: 'double') -> 'void':
+def initialize(ctx: 'void*') -> 'int':
```
williamjamir: Why is an `int` being returned?
Perhaps the docstring should be updated, with this information.

ggrbill: This is a ERROR CODE return. I will add a docstring. Thanks.

nicoddemus: Also include a description why the user would return something other than OK, and what the simulator will do when that happens. Also regarding `finalize`.

PR #14, opened 2018-12-03
Add new hook for secondary variable update on alfasim-sdk and API function for getting plugin variable from simulation

tarcisiofischer:
Add new hook for secondary variable update on alfasim-sdk and API function for getting plugin variable from simulation

ASIM-2346


tarcisiofischer:
mentioning @tadeu @arthursoprana just FYI

PR #15, opened 2018-12-20
Add hook compute_source_terms

ggrbill:


alfasim_sdk/hook_specs.py
```diff
@@ -22,6 +22,25 @@ def finalize(ctx: 'void*') -> 'int':
     :returns: Return OK if successful or anything different if failed
     """
 
+def compute_source_terms(
+    ctx: 'void*', 
+    mass_source: 'void*',
```
nicoddemus: Why are those parameters (other than the context) `void*`?

ggrbill: Because pybind cannot make the conversion of `double**` to array. It is need to make some conversion function, in which @tarcisiofischer is working at.

tarcisiofischer: Yeah, I couldn't solve this yet.
I asked help for pybind11 ppl (chat), but said this is not possible. Maybe we'll have to make wrapper functions, I don't know. :(

nicoddemus: So it is a 2D array... that should be in the description perhaps? Even mentioning which API functions to use to get the dimensions of the array

PR #18, opened 2019-01-22
Add alfasim_get_additional_variables

williamjamir:


alfasim_sdk/cli.py
```diff
@@ -43,9 +46,9 @@ def main():
 @click.option('--shared-lib-name', prompt='-- Shared Library Name', help='The filename of the compiled plugin')
 @click.option('--author-name', prompt='-- Author Name', help='Name of the plugin author to be displayed')
 @click.option('--author-email', prompt='-- Author Email', help='Email of the plugin author to be displayed')
-def template(dst, plugin_name, shared_lib_name, author_name, author_email, ):
+def template(dst, plugin_name, shared_lib_name, author_name, author_email,):
     """
-    Console script for alfasim_sdk.
+    Console script for ALFAsim_sdk.
```
nicoddemus: Should be kept as `alfasim_sdk` as it is the name of the module. 👍

alfasim_sdk/variables.py
```diff
@@ -0,0 +1,37 @@
...
+    Global = 'global'
+    Layer = 'layer'
+    Phase = 'phase'
+    # TODO ASIM - 2348 Add Scope.ContinuousField and Scope.DispersedField
+    # ContinuousField = 'continuous_field'
+    # DispersedField = 'dispersed_field'
+
+
+@attr.s(kw_only=True)
+class Variable():
```
nicoddemus: Perhaps we should depend on `barril` and validate the unit at this point?

alfasim_sdk/variables.py
```diff
@@ -0,0 +1,47 @@
...
```
tarcisiofischer: `Variable` is such a generic name. Perhaps we should call it at least `SecondaryVariable`, since in the future we may be able to also add `PrimaryVariable`s (I don't know yet how this would work, though). What do you think?

alfasim_sdk/variables.py
```diff
@@ -0,0 +1,47 @@
...
+
+import attr
+from attr import attrib
+from attr.validators import instance_of, optional
+from barril.units import UnitDatabase
+
+from alfasim_sdk._validators import check_for_valid_unit, check_string_is_not_empty
+
+
+class Visibility(Enum):
```
tarcisiofischer: Will those new classes have documentation? Or the user should look in an external document what does each one of those mean?

alfasim_sdk/_validators.py
```diff
@@ -0,0 +1,19 @@
...
+    """
+    if not value or value.isspace():
+        from alfasim_sdk.types import Enum
+        if type(instance) is Enum:
+            raise ValueError(f'Enum type cannot have an empty string on field "{attribute.name}"')
+
+        raise ValueError(f'The field "{attribute.name}" cannot be empty')
+
+
+def check_for_valid_unit(self, attribute, value):
```
nicoddemus: Here you are using `self`, in the previous method you used `instance`... I suggest using `self` to both to make it clear those methods are meant to be injected in classes later.

nicoddemus: For consistency with the above method (and because it reads better in English), I suggest to name this `check_unit_is_valid`.

alfasim_sdk/cli.py
```diff
@@ -43,7 +46,7 @@ def main():
 @click.option('--shared-lib-name', prompt='-- Shared Library Name', help='The filename of the compiled plugin')
 @click.option('--author-name', prompt='-- Author Name', help='Name of the plugin author to be displayed')
 @click.option('--author-email', prompt='-- Author Email', help='Email of the plugin author to be displayed')
-def template(dst, plugin_name, shared_lib_name, author_name, author_email, ):
+def template(dst, plugin_name, shared_lib_name, author_name, author_email,):
```
nicoddemus: This trailing `,` in the parameter list is weird

alfasim_sdk/hook_specs_gui.py
```diff
@@ -1,10 +1,17 @@
...
     """
-    Entry point for the creation of models in Alfasim
+    Entry point for the creation of models in ALFAsim
+    """
+
+
+@hookspec
+def alfasim_get_additional_variables():
+    """
+    Add additional variables in ALFAsim
```
nicoddemus: Improve this documentation a little bit, as it is it is very vague

alfasim_sdk/types.py
```diff
@@ -44,24 +46,27 @@ class String(BaseField):
...
-        if not isinstance(values, list):
-            raise TypeError(f"{attr.name} must be a list, got a {type(values)}.")
+    @values.validator
+    def check(self, attr, user_values):
+        if not isinstance(user_values, list):
+            raise TypeError(f"{attr.name} must be a list, got a {type(user_values)}.")
 
-        if not all(isinstance(value, str) for value in values):
+        if not all(isinstance(value, str) for value in user_values):
             raise TypeError(f"{attr.name} must be a list of string.")
```
nicoddemus: "list of strings"

Also include the offending item on the message. You might want to merge this check with th e`check_string_is_not_empty` below

alfasim_sdk/types.py
```diff
@@ -44,24 +46,27 @@ class String(BaseField):
...
-    value: List[str] = attrib()
+    values: List[str] = attrib()
     initial: str = attrib(validator=optional(instance_of(str)), default=None)
 
-    @value.validator
-    def check(self, attr, values):
-        if not isinstance(values, list):
-            raise TypeError(f"{attr.name} must be a list, got a {type(values)}.")
+    @values.validator
+    def check(self, attr, user_values):
```
nicoddemus: I'm curious why you changed `values` to `user_values`... `values` matches the name of the attribute we are validating after all. 🤔

alfasim_sdk/variables.py
nicoddemus: Both. 😁 

We need to include docstrings on those classes, which will be inserted in the manual documentation using `autodoc`.

alfasim_sdk/variables.py
nicoddemus: Sounds good.

tests/test_variables.py
```diff
@@ -0,0 +1,42 @@
+import pytest
+
+
+def test_additional_variable():
+    from alfasim_sdk.variables import Variable, Visibility, Location, Scope
+
+    with pytest.raises(ValueError, match='The field "name" cannot be empty'):
```
nicoddemus: You can turn all the `pytest.raises` check in a separate parametrized test, where you parametrize the attribute you want to change and the expected message.

```python
@pytest.mark.parametrize('attr, 'expected_message', [
    (dict(unit='foo'), "foo is not a valid unit"),
    (dict(caption='  '), "The field "caption" cannot be empty"),
    ...
])
def test_validation(attr, expected_message):
    attrs = dict(
        name='var_name',
        caption='Var Name',
        unit='m',
        visibility=Visibility.Output,
        location=Location.Center,
        multifield_scope=Scope.Global,
        checked_on_gui_default=False,
    )
    attrs.update(attr)
    with pytest.raises(TypeError, match=expected_message):
        Variable(**attrs)
```

tests/test_variables.py
```diff
@@ -0,0 +1,42 @@
...
+        visibility=Visibility.Output,
+        location=Location.Center,
+        multifield_scope=Scope.Global,
+        checked_on_gui_default=False,
+    )
+
+    assert variable.category == 'length'
+    assert variable.default_value == None
+
+    with pytest.raises(TypeError, match="'default_value' must be <class 'numbers.Real'>"):
```
nicoddemus: Why is this check far from the others?

alfasim_sdk/hook_specs_gui.py
williamjamir: What do you think of this docstring:
````python
@hookspec
def alfasim_get_additional_variables():
    """
    Allows plugins to register new additional variables on ALFAsim.
    This variable can be used to store internal data from the plugin,
    or it can be used to expose data to the user in the plot window.
    """
```

alfasim_sdk/types.py
williamjamir: What do you think?
```python
    with pytest.raises(TypeError, match="values must be a list of string, the item '1' is a 'int'"):
        Enum(values=[1], caption='caption')

File: alfasim_sdk/types.py (hunk 9 above)
Comment (williamjamir) [reply alfasim_sdk/types.py]: But it clashes with the decorator ` @values.validator` and I thought that this could be a problem.
Should I change?

File: tests/test_variables.py (hunk 11 above)
Comment (williamjamir) [reply tests/test_variables.py]: Fixed ;)

File: alfasim_sdk/_validators.py [hunk 12]
```diff
@@ -0,0 +1,19 @@
+def check_string_is_not_empty(instance, attribute, value):
```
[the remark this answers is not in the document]
williamjamir: The problem that I'm trying to solve here it's not the spaces within the string, but rather the possibility to have a caption empty ("") or a blank ("     ").

So in summary for me a caption:
❌ caption = ""
❌ caption = "   "
✔️ caption = "FOO"
✔️ caption = "FOO  "
✔️ caption = "   FOO  "
✔️ caption = "F  O  O  "


What do you think? Let me know about your thoughts on this check

alfasim_sdk/types.py
nicoddemus: Oh they definitely don't interfere with each other, please change it. :

alfasim_sdk/types.py
```diff
@@ -44,54 +58,65 @@ class String(BaseField):
...
+            raise TypeError(
+                f"{attr.name} must be a list, got a '{type(values).__name__}'."
+            )
 
-        if not all(isinstance(value, str) for value in values):
-            raise TypeError(f"{attr.name} must be a list of string.")
+        for value in values:
+            if not isinstance(value, str):
+                raise TypeError(
+                    f"{attr.name} must be a list of string, the item '{value}' is a '{type(value).__name__}'"
```
nicoddemus: typo: list of strings

alfasim_sdk/types.py
nicoddemus: Just a typo: "list of strings" ;)

alfasim_sdk/variables.py
```diff
@@ -7,10 +7,15 @@
...
 
-from alfasim_sdk._validators import check_for_valid_unit, check_string_is_not_empty
+from alfasim_sdk._validators import check_string_is_not_empty, check_unit_is_valid
 
 
 class Visibility(Enum):
+    """
+    Controls the visibility of the variable.
+        - Internal: The variable should only be used by the plugin, but not available to the end user.
+        - Output: The variable should be available to the end user, as a Property on Plot Window
```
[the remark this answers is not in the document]
williamjamir: This variable are not intended to be used by the end user; just by the plugin developer

alfasim_sdk/__init__.py
```diff
@@ -4,7 +4,7 @@
 import pluggy
 
 __author__ = """ESSS"""
-__email__ = 'foss@esss.co'
-__version__ = '0.1.0'
+__email__ = "foss@esss.co"
```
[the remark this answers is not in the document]
williamjamir: FOSS -> `Free and open-source software `

But if the question is why we decided to use this email.
A: I don't know 😁

alfasim_sdk/types.py
```diff
@@ -44,54 +58,65 @@ class String(BaseField):
...
-    caption = attrib(init=False, default='')
+    caption = attrib(init=False, default="")
 
-    def __attrs_post_init__(self):
+    def __attrs_post_init__(self) -> None:
         self.caption = self.value.caption
 
     @value.validator
-    def check(self, attr, values):
+    def check(self, attr: Attribute, values: Quantity) -> None:
```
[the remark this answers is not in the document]
williamjamir: @nicoddemus Can I disable this check " Parameters differ from overridden" ? 
The `check` method from `attrs` has the following signature: 
```
def check(self, attribute, value):
```
I've changed the `value` parameter to `values`

PR #20, opened 2019-02-06
Fb asim 2525 refactoring data reference

williamjamir:
1) Add a new type -> ReferenceSelection
2) Rename the class DataReference to Reference
3) Rename Reference.value to Reference.ref_type

### DIFF
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -99,6 +99,26 @@ class Reference(BaseField):
 
 
 @attr.s(kw_only=True)
+class ReferenceSelection(BaseField):
+    list_type = attrib()
+
+    @list_type.validator
+    def check(self, attr: Attribute, value: Type[ALFAsimType]) -> None:
+        if not isinstance(value, type):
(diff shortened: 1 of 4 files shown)


alfasim_sdk/types.py
```diff
@@ -89,15 +89,44 @@ def check(self, attr: Attribute, values: List[str]) -> None: # pylint: disable=a
...
+    """
+    ReferenceSelection allows the user to select multiples models as a "checklist" component.
+    This components only lists models that are Container Models "alfasim_sdk.models.container_models"
+
+    Properties:
+        caption   - property used as a label for the text input.
+        list_type - property that holds the container_model selected.
+
+    """
+    list_type = attrib()
```
williamjamir: @nicoddemus @tarcisioe And how bout about this name?

We call `list_type`,  but we actually just accept a single class that it's decorated by `@container_model`

nicoddemus: `container_type`?

PR #23, opened 2019-03-26
Add last hooks to hydrates

ggrbill:


alfasim_sdk/hook_specs.py
```diff
@@ -44,6 +44,53 @@ def compute_source_terms(
...
+def calculate_slurry_viscosity(
+    ctx: "void*",
+    alpha_f: "void*",
+    mu_f: "void*",
+    mu_f_layer: "void*",
+) -> "int":
+    """
+    Internal simulator hook to calculate slurry viscosity of layer(s).
+
+    :param ctx: ALFAsim's plugins context
```
nicoddemus: It is not clear which of those parameters are input and which are output...

alfasim_sdk/hook_specs.py
```diff
@@ -44,6 +44,53 @@ def compute_source_terms(
...
+    d_disp_fields: "void*",
+    P: "void*",
+    rho_f: "void*",
+    mu_f: "void*",
+    sin_theta_f: "void*",
+    delta_x_f: "void*",
+) -> "int":
+    """
+    Internal simulator hook to calculate slip velocity between fluids
+    and solids.
```
nicoddemus: It is not clear which of those parameters are input and which are output...

alfasim_sdk/hook_specs.py
ggrbill: I will add extra explanation to docstring

alfasim_sdk/hook_specs.py
ggrbill: I will add extra explanation to docstring

PR #26, opened 2019-04-10
Add new hook `update_plugins_secondary_variables_extra`

ggrbill:


alfasim_sdk/hook_specs.py
```diff
@@ -146,6 +146,18 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
...
 
+def update_plugins_secondary_variables_extra(ctx: "void*") -> "int":
+    """
+    Internal simulator hook to update plugin's secondary variables in the ExtraVarSolver scope.
+    ExtraVarsolver is used to solve the tracer transport equation.
+    This is called as the last step on ALFAsim's ExtraVarSolver update variables workflow.
+
+    :param ctx:
+
+    :returns: Return OK if successful or anything different if failed
```
williamjamir: This OK it's a macro?

alfasim_sdk/hook_specs.py
```diff
@@ -146,6 +146,18 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
     """
 
 
+def update_plugins_secondary_variables_extra(ctx: "void*") -> "int":
+    """
+    Internal simulator hook to update plugin's secondary variables in the ExtraVarSolver scope.
+    ExtraVarsolver is used to solve the tracer transport equation.
```
williamjamir: `ExtraVarSolver`

alfasim_sdk/hook_specs.py
```diff
@@ -146,6 +146,18 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
     """
 
 
+def update_plugins_secondary_variables_extra(ctx: "void*") -> "int":
+    """
+    Internal simulator hook to update plugin's secondary variables in the ExtraVarSolver scope.
```
williamjamir: Just a opinion, you can used it or not 😋 
```
Hook used internally by the simulator to update the plugins secondary variables on the ExtraVarSolver scope.

This hook will be executed right after the routine ExtraVarSolver has being finished, as the last step before continuing with the normal flow.

File: alfasim_sdk/hook_specs.py [hunk 4]
```diff
@@ -146,6 +146,18 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
     """
 
 
+def update_plugins_secondary_variables_extra(ctx: "void*") -> "int":
```
williamjamir: This hook's only used to update variables related with "tracers". Hence, I think that this hook should have a more meaningful name mentioning at least the tracers on his name.

Furthermore, we already have a hook named `update plugins secondary variables`, and adding a new hook just adding an `_extra` at the end could be confusing forcing the developer to explicitly look at the documentation to get the difference.

alfasim_sdk/hook_specs.py
ggrbill: It is an `enum` from `alfasim-sdk\alfasim_sdk_api\common.h`

alfasim_sdk/hook_specs.py
ggrbill: The second line is not what happens.

alfasim_sdk/hook_specs.py
ggrbill: But it is not only for tracers. The `ExtraVarSolver` can be used to solve extra variables, for instance, the moments from PBE of hydrates plugin.

PR #27, opened 2019-04-17
Fb asim 2744 variable getters

tarcisiofischer:


alfasim_sdk/hook_specs.py
```diff
@@ -146,6 +146,20 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
     """
 
 
+def update_plugins_secondary_variables_on_first_timestep(ctx: "void*") -> "int":
```
tarcisiofischer: @williamjamir asked to review all hook names. This will be done on a future PR, when we have more information about what the plugins actually need.

alfasim_sdk_api/detail/api_pointers.h
```diff
@@ -7,14 +7,15 @@ set_plugin_data_func set_plugin_data;
```
tarcisiofischer: IMO, API documentation should be moved to `api.h`, instead of leaving them on this detail header file, since the user shoudn't have to bother about the details. What do you guys think? @williamjamir @ggrbill

alfasim_sdk_api/common.h
```diff
@@ -1,8 +1,27 @@
...
+#define PHASE_GAS "gas"
+#define PHASE_LIQUID "liquid"
+#define PHASE_WATER "water"
+
+#define LAYER_GAS "gas"
+#define LAYER_LIQUID "liquid"
+#define LAYER_WATER "water"
+
+#define TIMESTEP_CURRENT 0
+#define TIMESTEP_OLD 1
```
ggrbill: Perhaps `CURRENT_TIMESTEP` and `PREVIOUS_TIMESTEP`. What do you think?

alfasim_sdk_api/detail/api_pointers.h
ggrbill: Agreed. It seems that things are hidden here.

alfasim_sdk_api/api.h
```diff
@@ -18,6 +18,33 @@ DLL_EXPORT int get_plugin_input_data_quantity(void* ctx, double* out, const char
...
 DLL_EXPORT int get_plugin_input_data_string_size(void* ctx, int* out, const char* plugin_name, const char* var_name);
 
-DLL_EXPORT int get_plugin_variable(void* ctx, void** out, const char* variable_name, int line_index, int* size);
+DLL_EXPORT int get_plugin_variable(void* ctx, void** out, const char* variable_name, int line_index, int timestep, int* size);
+
+DLL_EXPORT int get_field_id(void* ctx, int* out, const char* name);
+DLL_EXPORT int get_primary_field_id_of_phase(void* ctx, int* out, const char* name);
+DLL_EXPORT int get_phase_id(void* ctx, int* out, const char* name);
+DLL_EXPORT int get_layer_id(void* ctx, int* out, const char* name);
+DLL_EXPORT int get_simulation_array(
```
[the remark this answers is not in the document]
tarcisiofischer: Yes. It may change as we go. I tried to make a generic one to see how it behaves

alfasim_sdk_api/api.h
```diff
@@ -18,6 +18,33 @@ DLL_EXPORT int get_plugin_input_data_quantity(void* ctx, double* out, const char
...
+    void* ctx,
+    double** out,
+    char* variable_name,
+    int line_index,
+    int timestep,
+    int* size
+);
+DLL_EXPORT int get_simulation_quantity(
```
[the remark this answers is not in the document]
tarcisiofischer: Any non-array simulation variable.
The "quantity" concept is already stablished in our API. It has the same meaning from the other "quantities". To be more specific, it means a "double"

alfasim_sdk_api/common.h
tarcisiofischer: meh... Ok. Both look good to me.

ggrbill: OK, it is your decision. I just a suggested.

alfasim_sdk_api/detail/api_pointers.h
williamjamir: Yes, but when using the IDE, the "follow the implementation" or whatever other names it may be, brings the developer to this file.

alfasim_sdk_api/common.h
```diff
@@ -1,8 +1,52 @@
...
+    PHASE=3
+};
+
+enum TimestepScope
+{
+    CURRENT=0,
+    PREVIOUS=1
+};
+
+struct VariableScope
```
ggrbill: Nice!

alfasim_sdk_api/detail/api_pointers.h
tarcisiofischer: Ah... Ok. Thanks for pointing this out again.

PR #28, opened 2019-05-07
Add support for tab and tabs components

williamjamir:
ASIM-2778

### DIFF
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -2,11 +2,10 @@ import numbers
 from typing import Callable, List, Optional, Type, Union
 
 import attr
+from alfasim_sdk._validators import check_string_is_not_empty, check_unit_is_valid
 from attr import attrib
 from attr._make import Attribute
 from attr.validators import instance_of, optional
-
-from alfasim_sdk._validators import check_string_is_not_empty, check_unit_is_valid
 
 
 @attr.s(kw_only=True)
@@ -18,6 +17,21 @@ class ALFAsimType:
 class TracerType(ALFAsimType):
     type = attrib(default="tracer")
 
+
+@attr.s(kw_only=True)
+class Tab():
+    """
+    Base class for tab attributes available at ALFAsim.
+    """
+    pass
+
+
+@attr.s(kw_only=True)
+class Tabs():
+    """
+    Base class for tabs attributes available at ALFAsim.
+    """
+    pass
 
 @attr.s(kw_only=True)
 class BaseField:
@@ -58,12 +72,21 @@ class String(BaseField):
 
 
     """
-
     value: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
 
 
 @attr.s(kw_only=True)
 class Enum(BaseField):
+    """
+    The Enum field provides list of options to the user, showing  only the select item but providing a way to displays
+    list of all option trough a pop up.
+
+    Properties:
+        caption - A string to be displayed on the right side of the component.
+        values  - A list of strings to be added into the component as options to be selected.
+        initial - An optional property to indicate which one of the options should be selected per default.
+
+    """
     values: List[str] = attrib()
     initial: str = attrib(validator=optional(instance_of(str)), default=None)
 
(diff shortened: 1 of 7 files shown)


alfasim_sdk/types.py
```diff
@@ -19,6 +18,21 @@ class TracerType(ALFAsimType):
     type = attrib(default="tracer")
 
 
+@attr.s(kw_only=True)
+class Tab():
+    """
+    Base class for tab attributes available at ALFAsim.
+    """
+    pass
```
nicoddemus: Just FYI, you don't need those `pass` if you have a docstring in place. 👍

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
 
     """
-
     value: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
 
 
 @attr.s(kw_only=True)
 class Enum(BaseField):
+    """
+    The Enum field provides list of options to the user, showing  only the select item but providing a way to displays
```
```suggestion
    The Enum field provides list of options to the user, showing  only the select item but providing a way to display
```

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
+    list of all option trough a pop up.
```
```suggestion
    a list of all options through a combo-box.
```

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
+
+    Properties:
+        caption - A string to be displayed on the right side of the component.
```
```suggestion
        :ivar str caption: A string to be displayed on the right side of the component.
```

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
+        values  - A list of strings to be added into the component as options to be selected.
```
```suggestion
        :ivar List[str] values: A list of strings with the available options.
```

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
+        initial - An optional property to indicate which one of the options should be selected per default.
```
```suggestion
        :ivar str initial: Indicates which one of the options should be selected per default. If not given, the first item in ``values`` will be used as default.
```

alfasim_sdk/types.py
```diff
@@ -58,12 +72,21 @@ class String(BaseField):
...
     value: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
 
 
 @attr.s(kw_only=True)
 class Enum(BaseField):
+    """
+    The Enum field provides list of options to the user, showing  only the select item but providing a way to displays
+    list of all option trough a pop up.
+
+    Properties:
```
```suggestion

```

tests/test_layout.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_tabs():
+    from alfasim_sdk.types import String
+    from alfasim_sdk.layout import tabs, tab
+    from alfasim_sdk.models import data_model
+
+    error_msg = "Error on attribute 'a' expecting a class decorated with @tab but received a String type"
+    with pytest.raises(TypeError, match=error_msg):
+        @data_model(caption='Foo')
```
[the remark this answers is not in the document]
nicoddemus: Are you not using `A`?

alfasim_sdk/types.py
```diff
@@ -58,12 +70,20 @@ class String(BaseField):
...
     value: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
 
 
 @attr.s(kw_only=True)
 class Enum(BaseField):
+    """
+    The Enum field provides list of options to the user, showing  only the select item but providing a way to display
+    a list of all options through a combo-box.
+
+        :ivar str caption: A string to be displayed on the right side of the component.
```
```suggestion
    :ivar str caption: A string to be displayed on the right side of the component.
```

nicoddemus: You cannot indent it like this otherwise RST will interpret this as unformated text.

alfasim_sdk/types.py
```diff
@@ -58,12 +70,20 @@ class String(BaseField):
...
+        :ivar List[str] values: A list of strings with the available options.
```
```suggestion
    :ivar List[str] values: A list of strings with the available options.
```

alfasim_sdk/types.py
```diff
@@ -58,12 +70,20 @@ class String(BaseField):
...
+        :ivar str initial: Indicates which one of the options should be selected per default. If not given, the first item in ``values`` will be used as default.
```
```suggestion
    :ivar str initial: Indicates which one of the options should be selected per default. If not given, the first item in ``values`` will be used as default.
```

tests/test_layout.py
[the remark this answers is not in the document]
nicoddemus: But you were not referencing it anywhere?

PR #30, opened 2019-05-10
Add Int type to secondary variables of plugins

ggrbill:


alfasim_sdk/variables.py
```diff
@@ -41,6 +45,7 @@ class Scope(Enum):
 class SecondaryVariable:
     name: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
     caption: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
+    type = attrib(validator=instance_of(Type), default=Type.Double)
```
williamjamir: Update the test =)
Unfortunately, the Github doesn't allow me to add comments on files not change, but you basically you need to go to `test_variable.py` and a new entry on the parametrize =)

```
@pytest.mark.parametrize('attr, expected_type_error, expected_message', [
    (dict(name=''), ValueError, 'The field "name" cannot be empty'),
    (dict(caption=''), ValueError, 'The field "caption" cannot be empty'),
    (dict(unit='foo'), ValueError, 'foo is not a valid unit'),
    (dict(visibility=''), TypeError, "'visibility' must be <enum 'Visibility'>"),
    (dict(location=''), TypeError, "'location' must be <enum 'Location'>"),
    (dict(multifield_scope=''), TypeError, "'multifield_scope' must be <enum 'Scope'>"),
    (dict(default_value='1'), TypeError, "'default_value' must be <class 'numbers.Real'>"),
    (dict(type=''), TypeError, "'type' must be <enum "Type'>"),

])
def test_validation(attr, expected_type_error, expected_message):
    from alfasim_sdk.variables import Visibility, Location, Scope, SecondaryVariable
    attrs = dict(
        name='var_name',
        caption='Var Name',
        type=Type.Center
        unit='m',
        visibility=Visibility.Output,
        location=Location.Center,
        multifield_scope=Scope.Global,
        checked_on_gui_default=False,
    )
    attrs.update(attr)
    with pytest.raises(expected_type_error, match=expected_message):
        SecondaryVariable(**attrs)

williamjamir: Also, I think that you need to define a new name for this attribute, otherwise you will override the builtin [type](https://docs.python.org/3/library/functions.html#type) =)

ggrbill: Added, thanks!

PR #31, opened 2019-05-15
Add support for internal references on Reference field

williamjamir:
ASIM-2813

### DIFF
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -16,7 +16,7 @@ class ALFAsimType:
 
 @attr.s(kw_only=True)
 class TracerType(ALFAsimType):
-    container_type = attrib(default="TracerModelContainer")
+    _CONTAINER_TYPE = "TracerModelContainer"
 
 
 @attr.s(kw_only=True)
@@ -117,11 +117,11 @@ class Reference(BaseField):
 
     def __attrs_post_init__(self):
         if issubclass(self.ref_type, ALFAsimType):
-            self.container_type = attr.fields(self.ref_type).container_type.default
+            self.container_type = self.ref_type._CONTAINER_TYPE
         else:
             if self.container_type is None:
                 raise TypeError(
-                    f"The container_type field must be filled when ref_type is a class decorated with 'data_model'")
+                    f"The container_type field must be given when ref_type is a class decorated with 'data_model'")
 
     @ref_type.validator
     def check(self, attr: Attribute, value) -> None:
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -1,5 +1,5 @@
 import numbers
-from typing import Callable, List, Optional, Type, Union
+from typing import Callable, List, Optional, Union
 
 import attr
 from attr import attrib
@@ -16,7 +16,7 @@ class ALFAsimType:
 
 @attr.s(kw_only=True)
 class TracerType(ALFAsimType):
-    type = attrib(default="tracer")
+    container_type = attrib(default="TracerModelContainer")
 
 
 @attr.s(kw_only=True)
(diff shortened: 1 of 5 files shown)


alfasim_sdk/types.py
```diff
@@ -16,7 +16,7 @@ class ALFAsimType:
 
 @attr.s(kw_only=True)
 class TracerType(ALFAsimType):
```
ggrbill: What is the difference between `TracerType`, `TracerModel` and `TracerModelContainer`?

I am asking because I thought that we should have just `TracerModel` and `TracerModelContainer`.

alfasim_sdk/types.py
```diff
@@ -113,11 +113,28 @@ def check(self, attr: Attribute, values: List[str]) -> None:  # pylint: disable=
 @attr.s(kw_only=True)
 class Reference(BaseField):
     ref_type = attrib()
+    container_type = attrib(default=None, validator=[optional(instance_of(str)), optional(check_string_is_not_empty)])
+
+    def __attrs_post_init__(self):
+        if issubclass(self.ref_type, ALFAsimType):
+            self.container_type = attr.fields(self.ref_type).container_type.default
```
nicoddemus: Here check that `self.container_type is not None` and raise an error otherwise. 👍

alfasim_sdk/types.py
```diff
@@ -113,11 +113,28 @@ def check(self, attr: Attribute, values: List[str]) -> None:  # pylint: disable=
...
+        else:
+            if self.container_type is None:
+                raise TypeError(
+                    f"The container_type field must be filled when ref_type is a class decorated with 'data_model'")
```
```suggestion
                    f"The container_type field must be given when ref_type is a class decorated with 'data_model'")
```

alfasim_sdk/types.py
```diff
@@ -16,7 +16,7 @@ class ALFAsimType:
 
 @attr.s(kw_only=True)
 class TracerType(ALFAsimType):
-    type = attrib(default="tracer")
+    container_type = attrib(default="TracerModelContainer")
```
```suggestion
    _CONTAINER_TYPE = "TracerModelContainer"
```

This seems like it should be a class attribute instead.

alfasim_sdk/types.py
```suggestion
            self.container_type = self.ref_type._CONTAINER_TYPE
```

Following my comment about this being a class attribute.

tests/test_types.py
```diff
@@ -50,37 +50,74 @@ def test_enum():
...
-    with pytest.raises(TypeError, match="arg 1 must be a class"):
+    with pytest.raises(TypeError, match='ref_type must be a class'):
         Reference(ref_type="", caption="caption")
 
-    with pytest.raises(TypeError, match="ref_type must be a valid ALFAsim type"):
-        Reference(ref_type=Data1, caption="caption")
+    with pytest.raises(TypeError, match="ref_type must be an ALFAsim type or a class decorated with 'data_model'"):
+        Reference(ref_type=InvalidClass, caption="caption")
+
+    error_msg = "ref_type must be an ALFAsim type or a class decorated with 'data_model', got a class decorated with 'container_model'"
```
nicoddemus: If possible, move those checks to a function and reuse them in the MultipleReference test

alfasim_sdk/types.py
```diff
@@ -132,16 +149,29 @@ class MultipleReference(BaseField):
...
+            if self.container_type is None:
+                raise TypeError(
+                    f"The container_type field must be filled when ref_type is a class decorated with 'container_model'")
 
-    @container_type.validator
-    def check(self, attr: Attribute, value: Type[ALFAsimType]) -> None:
-        if not isinstance(value, type):
-            raise TypeError(f"{attr.name} must be a class, got {type(value).__name__}")
+    @ref_type.validator
+    def check(self, attr: Attribute, value) -> None:
```
oliveira-esss: I think this whole validator is duplicated from lines 127-137

alfasim_sdk/types.py
```diff
@@ -132,16 +149,29 @@ class MultipleReference(BaseField):
...
+        if type(value) is not type:
+            raise TypeError(f"{attr.name} must be a class")
 
         if not issubclass(value, ALFAsimType):
-            raise TypeError(f"{attr.name} must be a valid ALFAsimType, got {type(value).__name__}")
+            if not hasattr(value, "_alfasim_metadata"):
+                raise TypeError(f"{attr.name} must be an ALFAsim type or a class decorated with 'data_model'")
+
+            if value._alfasim_metadata["model"] is not None:
```
oliveira-esss: Hmm having a value different than None in this attribute means it is a container_model decorated class? Is that supposed to be obvious and I'm missing something? If not, what do you think about adding a comment like "data model classes will have model set to None"

alfasim_sdk/types.py
williamjamir: Sorry I didn't catch what you mean it by `check that is not None` and raise if it's `None`

```python
if self.container_type is not None:
    raise TypeError("A container_type should not be given when an ALFAsimType is given)
self.container_type = self.ref_type._CONTAINER_TYPE
```
or
```python
self.container_type = self.ref_type._CONTAINER_TYPE
if self.container_type is None:
    raise Type("Something went wrong, did you forgot to specify _CONTAINER_TYPE?")

File: alfasim_sdk/types.py (hunk 1 above)
Comment (williamjamir) [reply alfasim_sdk/types.py]: Only the `TracerType` will be used ;)
`container_type` and is internally only, and the user will not be in contact with these values.

File: alfasim_sdk/types.py (hunk 7 above)
Comment (williamjamir) [reply alfasim_sdk/types.py]: Basically, the difference between `data_model` and `container_model` is the "model attribute".

I would not say that it's an "obvious" statement, but this is a behavior that we are following since the beginning. Note that there isn't any enforcement on `container_model` perhaps we should 😬 

The docstring from `data_model` and `container_model` describes try to explain this behavior.

From the `container_model`
```python
1) Application Required:
    All properties that are required from the application can be accessed from _alfasim_metadata.

    Currently, the following options are available:
        - caption: A text to be displayed
        - icon: Name of the icon available at resource folder to be used over the TreeStructure
        - model: A reference for a user-declared class that has the @data_model decorator.
```

from the `data_model`
```python
 1) Application Required:
    All properties that are required from the application can be accessed from _alfasim_metadata.

    Currently, the following options are available:
        - caption: A text to be displayed
        - icon: Name of the icon available at resource folder to be used over the TreeStructure
        - model: None (data_model cannot reference another model).
```

oliveira-esss: Hmm ok. The docstrings are pretty clear, but I still would add some quick doc close to that code.

PR #32, opened 2019-05-23
Add get_plugin_input_data_table_quantity API function to get table co…

tarcisiofischer:
…lumn data from plugins

ASIM-2523

### DIFF
--- a/alfasim_sdk_api/detail/api_pointers.h
+++ b/alfasim_sdk_api/detail/api_pointers.h
@@ -60,6 +60,41 @@ typedef int (*get_plugin_input_data_string_size_func)(void*, int*, const char*, 
 get_plugin_input_data_string_size_func get_plugin_input_data_string_size;
 
 /**
+*   get_plugin_input_data_table_quantity
+*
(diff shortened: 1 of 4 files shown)


alfasim_sdk_api/detail/api_pointers.h
```diff
@@ -59,6 +59,37 @@ get_plugin_input_data_string_func get_plugin_input_data_string;
...
+*   int size = -1;
+*   double* values = NULL;
+*   int errcode = get_plugin_input_data_table_quantity(
+*       ctx,
+*       &values,
+*       &size,
+*       "temperature",
+*       plugin_name,
+*       "some_table"
+*    );
```
nicoddemus: Would be nice to show an example of how to actually use `values` (together with `size` of course)`

tarcisiofischer: Sure

tarcisiofischer: Done.

PR #33, opened 2019-05-24
Add new API function to access tracers' mass fraction

ggrbill:


alfasim_sdk_api/detail/api_pointers.h
```diff
@@ -129,6 +129,34 @@ typedef int (*get_simulation_array_func)(
 );
 get_simulation_array_func get_simulation_array;
 
+/**
+*   get_simulation_tracer_array
+*
+*   Get the current contents of a given tracer mass fraction (For an array data pointer).
+*   A  tracer mass fraction is calculated in the extra solver iterative procedure.
```
tarcisiofischer: This docs seem OK for now, but we'll probably review them later - I'm unsure if we'll expose this much information about the solver and the existence of the extra solver (Maybe we will - I'm just saying we may review this later)

ggrbill: Sure. I agree. In fact, I think we should review soon the API related to `tracers`.

PR #34, opened 2019-05-27
Add GroupFied on ALFAsim SDK

williamjamir:


alfasim_sdk/_alfasim_sdk_utils.py
```diff
@@ -1,8 +1,7 @@
 from typing import Optional, Tuple
 
 import attr
-
-from alfasim_sdk.types import BaseField, Tab, Tabs
+from alfasim_sdk.types import BaseField, Tab, Tabs, Group
```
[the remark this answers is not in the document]
williamjamir: Commit 983e1d0 fix this issue =)

williamjamir:
> Why so many formatting changes?

I executed `ff` and `black` together: mild-panic:

I'm going to revert the commit and open a separate PR adding a code formatting check on CI ;)

tests/test_layout.py
```diff
@@ -1,4 +1,64 @@
...
+
+
+        @group(caption='This is a Group')
+        class GroupMain:
+            string_from_group = String(caption='Caption From Group', value='Inside Group')
+
+
+    import attr
+    # Checking the Model has all attributes
+    assert len(attr.fields(ValidClass)) == 2
```
nicoddemus: You can replace all those 3 lines with:

```python
assert [x.name for x in attr.fields(ValidClass)] == ['string_from_main', 'GroupMain']
```

This also provides a better diagnosis when it fails. 😉

tests/test_layout.py
```diff
@@ -1,4 +1,64 @@
...
+    assert attr.fields(ValidClass)[0].name == 'string_from_main'
+    assert attr.fields(ValidClass)[1].name == 'GroupMain'
+
+    # Checking attributes from the GroupMain
+    group_class = attr.fields(ValidClass)[1].default
+    assert len(attr.fields(group_class)) == 1
```
nicoddemus: Same here

```python
assert [x.name for x in attr.fields(group_class)] == ["string_from_group"]
```

alfasim_sdk/_alfasim_sdk_utils.py
```diff
@@ -2,7 +2,7 @@
 
 import attr
 
-from alfasim_sdk.types import BaseField, Tab, Tabs
+from alfasim_sdk.types import BaseField, Group, Tab, Tabs
 
 
 def get_attr_class(class_: type, caption: str, icon: Optional[str], model: Optional[type], bases: Optional[Tuple]=()):
```
prusse-martin: `Optional` type hint is for argument that accept `None` value and not those optional due having a default value.
Maybe it has a non `None` default but will accept `None` then this is just a comment =)

williamjamir: I once read on mypy cheat sheet that is a good practice to provide type annotation to variables that initialized with "None" =)

I know that the instructions clearly say variables, but I think that it's a good argument to use it on  function arguments :stuck_out_tongue_closed_eyes: 
[Link for the cheat_ sheet](https://mypy.readthedocs.io/en/latest/cheat_sheet_py3.html#when-you-re-puzzled-or-when-things-are-complicated)

PR #35, opened 2019-05-28
Add codeformatter

williamjamir:


CONTRIBUTING.rst
```diff
@@ -7,122 +7,69 @@ Contributing
...
-* Your operating system name and version.
-* Any details about your local setup that might be helpful in troubleshooting.
-* Detailed steps to reproduce the bug.
-
-Fix Bugs
-~~~~~~~~
-
-Look through the GitHub issues for bugs. Anything tagged with "bug" and "help
-wanted" is open to whoever wants to implement it.
+Get Started!
```
nicoddemus: We should add a blurb about how to develop `alfasim_sdk` inside ESSS.

alfasim_sdk/variables.py
```diff
@@ -3,19 +3,19 @@
 from typing import Optional
 
 import attr
+from alfasim_sdk._validators import check_string_is_not_empty, check_unit_is_valid
```
nicoddemus: This import seems strange, should be kept alone...

nicoddemus: Another idea is that we can try using https://github.com/asottile/reorder_python_imports instead, which should present fewer problems and has native pre-commit integration. 👍

williamjamir: Sorry, I saw your comments just after adding the `seed-isort-config` on pre-commit, the next commit will add your suggestion ;)

CONTRIBUTING.rst
williamjamir: Commit https://github.com/ESSS/alfasim-sdk/pull/35/commits/0ee145fe31cba6ec5ee22e099ea0fd124f877976 add some instructions, could review it?

williamjamir:
How do I create the environment?

As usual, the workflow didn't change, the only think nows it's that the Developer needs to execute `pre-commit install` from inside the alfasim-sdk repository to enable the git commit hook.commit.

> Does the command below work in my apps environment as usual (for example, is pre-commit installed by qmxgraph's environment?)

So, yes and no.
If you have the app environment activate everything will work smoothly, because he has already the dependency installed on the requirements. (but again, the developer needs to explicitly install the pre-commit on the alfasim-SDK once).

If I only activate the environment from alfasim-sdk this will not work (because pre-commit it's not an explicit dependency), but I don't see a "user case" where someone from inside the ESSS will commit something from inside an "alfasim-SDK" environment.

nicoddemus:
Thanks for the clarification, please add those observations to the docs. 👍

williamjamir:
> Thanks for the clarification, please add those observations to the docs. 👍

Done, I've added right at the beginning of the document this instruction and create a section FAQ for developing from inside ESSS.

Maybe a little too much? 😀

PR #41, opened 2019-06-10
Implement plugin status monitor

williamjamir:


alfasim_sdk/status.py
```diff
@@ -0,0 +1,17 @@
+import attr
+from attr import attrib
+from attr.validators import instance_of
+
+from alfasim_sdk._validators import check_string_is_not_empty
+
+
+@attr.s(kw_only=True)
+class ErrorMessage:
+    model_name: str = attrib(validator=[instance_of(str), check_string_is_not_empty])
```
nicoddemus: `check_string_is_not_empty` should already use `instance_of(str)` behind the scenes, so users can simply use:

```python
model_name: str = attrib(validator=check_string_is_not_empty)
```

tests/test_status.py
```diff
@@ -0,0 +1,29 @@
+import pytest
+
+from alfasim_sdk.status import ErrorMessage
+from alfasim_sdk.status import WarningMessage
+
+
+@pytest.mark.parametrize("status", [WarningMessage, ErrorMessage])
```
nicoddemus: I suggest using `status_class`

tests/test_status.py
```diff
@@ -0,0 +1,29 @@
...
+def test_status(status):
+    out = status(model_name="Test1", message="Message From test")
+    assert out.model_name == "Test1"
+    assert out.message == "Message From test"
+
+    with pytest.raises(ValueError, match='The field "model_name" cannot be empty'):
```
nicoddemus: TBH I don't see much point in this test, you are basically testing `attr.s` functionality...

alfasim_sdk/hook_specs_gui.py
```diff
@@ -19,6 +19,15 @@ def alfasim_get_additional_variables():
     """
 
 
+@hookspec
+def alfasim_get_status(ctx):
+    """
+    Allows plugins to execute custom checks on ALFAsim.
```
nicoddemus: One of the most important things we should document is inputs and expected outputs... feel free to use type annotations here.

tests/test_status.py
williamjamir: Yep, basically, I'm just testing that the declaration has a validator and no one has removed/overwritten by mistake or something similar.

williamjamir:
> Just realized that a better name is `non_empty_str`, as it better matches the other validators (a noun rather than a verb)

Yep, `non_empty_str` sounds way better =)

What do you think about to change `check_unit_is_valid` to `valid_unit`?

nicoddemus:
> What do you think about to change check_unit_is_valid to valid_unit?

Sounds good!

tests/test_types.py
```diff
@@ -195,7 +195,7 @@ def test_tooltips():
     field = Boolean(value=True, caption="caption", tooltip="Test123")
     assert field.tooltip is "Test123"
 
-    expected_msg = "'tooltip' must be <class 'str'> \(got 2 that is a <class 'int'>\)."
+    expected_msg = r"'tooltip' must be <class 'str'> \(got 2 that is a <class 'int'>\)."
```
nicoddemus:
```suggestion
    expected_msg = re.escape(r"'tooltip' must be <class 'str'> (got 2 that is a <class 'int'>).")
```

williamjamir: Commit 00b3a1b applies your suggestion ;)

nicoddemus:
Good work!

PR #48, opened 2019-06-28
Add FilePath type

williamjamir:
This pull request adds support for FilePath type

### DIFF
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -248,3 +248,19 @@ class Table(BaseField):
 @attr.s(kw_only=True)
 class Boolean(BaseField):
     value: bool = attrib(validator=instance_of(bool))
+
+
+@attr.s(kw_only=True)
+class FilePath(BaseField):
+    """
+    The FilePath component provides a platform-native file dialog to the user to be able to select a file.
+    The name of the selected file will be available over the GUI e be enabled to be manually changed.
+    Be aware that the provided file from the user could be wrong, or not available when the application
+    is closed and opened again, therefore, is recommended to include a status monitor in your plugin
+    to make sure that the file selected is valid and has the desired format.
+
+    For more details about status monitor check alfasim_sdk.status.ErrorMessage
+
+    Properties:
+        caption - label to be used on the left side of the component, that informs the selected file.
+    """
--- a/alfasim_sdk/types.py
+++ b/alfasim_sdk/types.py
@@ -1,5 +1,6 @@
 import numbers
 from typing import Callable
+from typing import FrozenSet
 from typing import List
 from typing import Optional
 from typing import Union
@@ -204,20 +205,20 @@ class MultipleReference(BaseReference):
     """
 
 
-@attr.s(kw_only=True)
+@attr.s(kw_only=True, frozen=True)
 class Quantity(BaseField):
     value: numbers.Real = attrib(validator=instance_of(numbers.Real))
     unit: str = attrib(validator=[non_empty_str, valid_unit])
 
 
-@attr.s(kw_only=True)
+@attr.s(kw_only=True, frozen=True)
 class TableColumn(BaseField):
     id: str = attrib(validator=non_empty_str)
     value: Quantity = attrib()
     caption = attrib(init=False, default="")
 
     def __attrs_post_init__(self) -> None:
-        self.caption = self.value.caption
+        object.__setattr__(self, "caption", self.value.caption)
 
     @value.validator
     def check(  # pylint: disable=arguments-differ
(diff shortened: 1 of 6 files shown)


alfasim_sdk/types.py
```diff
@@ -248,3 +248,19 @@ def check(  # pylint: disable=arguments-differ
 @attr.s(kw_only=True)
 class Boolean(BaseField):
     value: bool = attrib(validator=instance_of(bool))
+
+
+@attr.s(kw_only=True)
+class FilePath(BaseField):
+    """
+    The FilePath component provides a platform-native file dialog to the user to be able to select a file.
+    The name of the selected file will be available over the GUI e be enabled to be manually changed.
```
nicoddemus: `e` -> `and`

alfasim_sdk/types.py
```diff
@@ -248,3 +248,19 @@ def check(  # pylint: disable=arguments-differ
...
+    Be aware that the provided file from the user could be wrong, or not available when the application
```
nicoddemus: "wrong"? what do you mean?

alfasim_sdk/types.py
```diff
@@ -248,3 +248,19 @@ def check(  # pylint: disable=arguments-differ
...
+    is closed and opened again, therefore, is recommended to include a status monitor in your plugin
+    to make sure that the file selected is valid and has the desired format.
+
+    For more details about status monitor check alfasim_sdk.status.ErrorMessage
+
+    Properties:
+        caption - label to be used on the left side of the component, that informs the selected file.
```
```rst
:ivar caption: label to be used on the left side of the component, that informs the selected file.
```

Let's use proper markup so when we get to generate the docs we have less work to do. 👍

tests/test_types.py
```diff
@@ -188,6 +188,13 @@ def test_boolean():
         Boolean(value=1, caption="caption")
 
 
+def test_filepath():
+    from alfasim_sdk.types import FilePath
+
+    file_path = FilePath(caption="Test")
+    assert file_path
```
nicoddemus: For a smoke test only creating the object is enough; `assert file_path` gives the impression that it might implement `__bool__`, which is not the case. Thus I suggest to use simply:

```python
def test_filepath():
    from alfasim_sdk.types import FilePath

    FilePath(caption="Test")    
```

alfasim_sdk/types.py
prusse-martin: If the project file is shared the sharing user should be aware that another file should be also shared and the destination user that this other file must be placed, possibly, in another folder?

I don't think this is a good way to tackle this.
Why not load this data into memory then dump that to the project file?
Or at least copy the file to a known an easy to share location?

nicoddemus: Definitely, thanks for bringing this to attention.

As we discussed, we should copy the file to the data directory of the project, and the plugin accesses that file path. 👍

tarcisiofischer: Wow. Could this at least be optional? I mean, if the file is really big (Imagine a 9999GB read-only file), this means that the user is doomed to wait the file copy to always happen?
Also, if the file comes from a read-only server where the user cannot copy stuff from, he may even be unable to use the plugin.

I think assuming the file is available and checking it from the status monitor is actually reasonable, TBH.
Am I thinking too hard here?

nicoddemus: >  I mean, if the file is really big (Imagine a 9999GB read-only file)

Come on... example by absurd? 😛 

Our usual case is files of reasonable sizes, not Google-size index databases.

> Also, if the file comes from a read-only server where the user cannot copy stuff from, he may even be unable to use the plugin.

And how he would use the file at all, if he can't read it? 

> I think assuming the file is available and checking it from the status monitor is actually reasonable, TBH.

Not really, because if the user transfers the file to another user, the *full path* will be saved, which is most certainly wrong.

> Am I thinking too hard here?

Perhaps. 😁 

We will be talking about files of a few mega bytes in almost all cases. The safe bet is to just copy the file next to the project. It also allows easy sharing of the project file/directory without nasty surprises.

tarcisiofischer: Fair enough

alfasim_sdk/types.py
```diff
@@ -227,15 +230,15 @@ def check(  # pylint: disable=arguments-differ
...
 class Table(BaseField):
-    rows: List[TableColumn] = attrib()
+    rows: FrozenSet[TableColumn] = attrib(converter=frozenset)
 
     @rows.validator
     def check(  # pylint: disable=arguments-differ
         self, attr: Attribute, values: Union[List[str], str]
     ):
-        if not isinstance(values, list):
+        if not isinstance(values, frozenset):
```
nicoddemus: Doesn't the converter ensure this will always be true?

williamjamir: Yep, this check is redundant I will remove it =)

tests/test_types.py
nicoddemus: Please remove this `assert` 👍

alfasim_sdk/types.py
```diff
@@ -248,3 +246,18 @@ def check(  # pylint: disable=arguments-differ
...
+class FilePath(BaseField):
+    """
+    The FilePath component provides a platform-native file dialog to the user to be able to select a file.
+    The name of the selected file will be available over the GUI and be enabled to be manually changed.
+
+    If you want to make the file mandatory is recommended to include a status monitor in your plugin
+    to make sure that that a file is selected.
+
+    For more details about status monitor check alfasim_sdk.status.ErrorMessage
+    Properties:
```
nicoddemus: Remove `Properties` and dedent `:ivar caption` please

tests/test_types.py
williamjamir: Solved ;)

alfasim_sdk/types.py
nicoddemus: Hmm this is still here 😕

williamjamir: I solved the conversation before the push 😄

PR #51, opened 2019-07-10
Add multiple-reference related API functions

tarcisiofischer:
ASIM-2866

### DIFF
--- a/alfasim_sdk_api/detail/api_pointers.h
+++ b/alfasim_sdk_api/detail/api_pointers.h
@@ -373,38 +373,29 @@ set_wall_layer_property_func set_wall_layer_property;
 
 
 /**
-*   get_plugin_input_data_multiplereference_selected_indexes_size
+*   get_plugin_input_data_multiplereference_selected_size
 *
-*   Get the number of selected references in a multiple-reference selection. To be used with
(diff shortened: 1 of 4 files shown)


alfasim_sdk_api/detail/api_pointers.h
```diff
@@ -371,4 +371,40 @@ get_wall_layer_id_func get_wall_layer_id;
...
+*
+*   Get the index of the selected items in a multiple-reference selection. To get the array size,
+*   you should use get_plugin_input_data_multiplereference_selected_indexes_size.
+*
+*   Example usage:
+*   int errcode = -1;
+*
+*   int indexes_size = -1;
+*   errcode = get_plugin_input_data_multiplereference_selected_indexes_size(ctx, &indexes_size, "Model.multiple_reference");
+*   int* indexes = (int*) malloc(sizeof(int) * indexes_size);
```
nicoddemus: If we are using C++ on the examples (`std::string` below), I think we should stick to it thoroughly.

```c++
std::vector<int> indexes(indexes_size);
```

tarcisiofischer: This code will change, as we decided that it's not really necessary

PR #57, opened 2019-08-08
Refactor on plugin config 

williamjamir:


prusse-martin:
> Again, as I said in the `hookman` changes PR, this `id` sounds like a number, but it is not. For me, it would be more clear if we had `caption` and `name` instead of `caption` and `id`.

Can i cite wikipedia? I will cite wikipedia =)
https://en.wikipedia.org/wiki/Identifier
> The abbreviation ID often refers to identity, identification (the process of identifying), or an identifier (that is, an instance of identification). An identifier may be a word, number, letter, symbol, or any combination of those. 

IMO we just need to document stuff.

PR #64, opened 2019-09-09
Set dll handle to nullptr after close the shared library

ggrbill:
It can cause an access violation in a test that runs after another test
that failed and the dll handle wasn't cleaned setting nullptr


nicoddemus:
The failure is unrelated, but we definitely should investigate and fix it. @williamjamir can you take a look perhaps?

PR #72, opened 2019-10-02
Add Poetry

williamjamir:


tox.ini
```diff
@@ -1,22 +1,29 @@
...
+    poetry run codecov -t CODECOV_TOKEN
 
 [testenv:linting]
 skip_install = True
 basepython = python3.6
 deps = pre-commit>=1.11.0
 commands = pre-commit run --all-files --show-diff-on-failure
+
+# [tox:.package]
+# # note tox will use the same python version as under what tox is installed to package
```
nicoddemus: Is this required?

williamjamir: I think that it's not necessary, even though the tox documentation recommends it (perhaps I didn't understand the docs correctly)

[Link for the section on docs documentation that mentions poetry](https://tox.readthedocs.io/en/latest/example/package.html#poetry)

nicoddemus: It is not really needed, we only support python3 anyway... please remove it. 👍

PR #84, opened 2019-12-13
Add minimum alfasim sdk required version to plugins ASIM-3031

oliveira-esss:


alfasim_sdk/_alfasim_sdk_utils.py
```diff
@@ -72,3 +72,26 @@ def get_all_attributes(class_: type):
...
+
+
+def get_extras_default_required_version():
+    """
+    :rtype str:
+    :return:
+        Returns a string with default alfasim-sdk version requirement for plugins. Default is
+        greater or equal current version and lesser than next major release.
+    """
+    current_version = get_current_version()
```
nicoddemus: For strict correctness we should require >= current minor, so bug fixes are not part of the required version:

```python
    parts = current_version.split(".")
    current_major = parts[0]
    current_minor = ".".join(parts[:2])
    next_major = int(current_major) + 1
    return f">={current_minor}, <{next_major}"
```

alfasim_sdk/cli.py
```diff
@@ -61,7 +61,7 @@ def template(dst, caption, plugin_id, author_name, author_email):
     The template folder will be placed on the ``dst`` option, that by default is the current directory from where the command
```
nicoddemus: We need to change this docstring into a raw string to fix the build, change line 58 from `"""` to `r"""`

alfasim_sdk/__init__.py
```diff
@@ -4,7 +4,6 @@
 
 __author__ = """ESSS"""
 __email__ = "foss@esss.co"
-__version__ = "0.1.0"
```
nicoddemus: Why was this removed?

oliveira-esss: Version is already in `pyproject.toml`

PR #86, opened 2020-02-17
Move HydrodynamicModelType to constants.py and Add new entries

williamjamir:
ASIM-3477

### DIFF
--- a/alfasim_sdk/constants.py
+++ b/alfasim_sdk/constants.py
@@ -58,35 +58,23 @@ class HydrodynamicModelType(Enum):
     - FourFields - 'Multi-field, Unit Cell (gas-liquid)':
         Two phase (gas and liquid) with four fields (continuous gas, continuous liquid, dispersed gas bubble, and dispersed liquid droplet).
 
-    - FiveFieldsSolid
-        ...
-
-    - FiveFieldsWater
-        ...
-
-    - FiveFieldsCO2
-        ...
-
     - ThreeLayersGasOilWater - 'Multi-field, Unit Cell (gas-liquid-water)':
         Three phase (gas, liquid, and water) with five fields (continuous gas, continuous liquid, continuous water, dispersed gas bubble, and dispersed liquid droplet).
 
-    - ThreeLayersNoBubbleGasOilWater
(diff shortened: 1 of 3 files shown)


alfasim_sdk/constants.py
```diff
@@ -42,3 +44,50 @@
 """
 The dict key which identifies the required version of alfasim-sdk for a given plugin
 """
+
+
+class HydrodynamicModelType(Enum):
```
williamjamir: @arthursoprana, @ggrbill , @sofiapbittencourt , @edersongrein 

Any suggestions for the docstring of the following items?
- FiveFieldsSolid
- FiveFieldsWater
- FiveFieldsCO2
- ThreeLayersNoBubbleGasOilWater
- ThreeLayersWaterWithCO2

Just remembering that this content of this docstring goes directly to the ALFAsim-SDK documentation.

alfasim_sdk/constants.py
```diff
@@ -42,3 +44,50 @@
...
+
+    TwoFields is valid only for the slug/regime capturing
+
+    - TwoFields - 'Two-fluid, Regime Capturing (gas-liquid)':
+        Two phase (gas and liquid) with two fields (continuous gas and continuous liquid) using Regime Capturing strategy.
+
+    - FourFields - 'Multi-field, Unit Cell (gas-liquid)':
+        Two phase (gas and liquid) with four fields (continuous gas, continuous liquid, dispersed gas bubble, and dispersed liquid droplet).
+
+    - FiveFieldsSolid
```
ggrbill: Please, remove those types that are not available at GUI of ALFAsim.

alfasim_sdk/constants.py
ggrbill: I suggest to remove just from the docstrings and put some comment in the code saying that those types are not supported (since they are not available on GUI), or list them at the end of docstring also saying that they are not supported.

arthursoprana: As @nicoddemus suggested in Rocket.Chat, add a `Internal` or `Do not use` tag.

PR #87, opened 2020-02-19
Add new hook for update internal deposit layer

sofiapbittencourt:
ASIM-3521

### DIFF
--- a/alfasim_sdk/hook_specs.py
+++ b/alfasim_sdk/hook_specs.py
@@ -331,9 +331,9 @@ def calculate_mass_source_term(
                 (double*) mass_source + n_control_volumes * liq_id;
             // Make some calculations and add it to liquid_mass_source.
             // In this example, we add a mass source of 3.1415 kg/s to all control volumes.
-			for (int i = 0; i < n_control_volumes; ++i) {
-    			liquid_mass_source[i] = 3.1415; // [kg/s]
-			}
+            for (int i = 0; i < n_control_volumes; ++i) {
+                liquid_mass_source[i] = 3.1415; // [kg/s]
+            }
             return OK;
         }
 
@@ -1333,6 +1333,52 @@ def calculate_entrained_liquid_fraction(
     """
 
 
+def update_plugins_internal_deposit_layer(ctx: "void*", deposition_layer_thickness: "void*",
+                                             n_control_volumes: "int") -> "int":
+    """
+    **c++ signature** : ``HOOK_UPDATE_PLUGINS_INTERNAL_DEPOSIT_LAYER(void* ctx, void* deposition_layer_thickness,
+    int n_control_volumes)``
+
+    Internal simulator hook to evaluate the thickness of the deposited layer at the inside of the pipeline walls.
+    This is called at the beginning of accounting the diameter reduction.
+
+    The plugin is supposed to change the given ``deposition_layer_thickness`` array pointer. Its values are contiguous
+    in memory and the dimension is given by ``n_control_volumes``. It has unit equal to ``[m]``.
+
+    :param ctx: ALFAsim's plugins context
+    :param deposition_layer_thickness: Thickness of the internal deposition layer
+    :param n_control_volumes: Number of control volumes
+    :returns: Return OK if successful or anything different if failed
+
+    Example of usage:
+    TODO!!!!!!!!
+    .. code-block:: c++
+        :linenos:
+        :emphasize-lines: 1
+
+        HOOK_UPDATE_PLUGINS_INTERNAL_DEPOSIT_LAYER(
+            ctx, deposition_layer_thickness, n_control_volumes)
+        {
+            errcode = alfasim_sdk_api.get_field_id(
+                ctx, &liquid_id, "liquid");
+            // Convertion from void* to double* and getting the
+            // array range related to liquid field
+            double* liquid_mass_source =
+                (double*) mass_source + n_control_volumes * liq_id;
+            // Make some calculations and add it to liquid_mass_source.
+            // In this example, we add a mass source of 3.1415 kg/s to all control volumes.
+            for (int i = 0; i < n_control_volumes; ++i) {
+                liquid_mass_source[i] = 3.1415; // [kg/s]
+            }
+            return OK;
+    }
+
+    In the example above is shown how to manage the ``mass_source`` array to get the mass source term array related to a
+    specific field (`liquid field` in this case). Note that ``liquid_mass_source`` has size equal to ``n_control_volumes``.
+
+    """
+
+
 specs = HookSpecs(
     project_name="ALFAsim",
     version="1",
@@ -1369,5 +1415,7 @@ specs = HookSpecs(
         friction_factor,
         env_temperature,
         calculate_entrained_liquid_fraction,
+        # Internal Deposition Layer
+        update_plugins_internal_deposit_layer,
     ],
 )
--- a/alfasim_sdk/hook_specs.py
+++ b/alfasim_sdk/hook_specs.py
@@ -1333,10 +1333,11 @@ def calculate_entrained_liquid_fraction(
     """
 
 
-def update_plugins_internal_deposit_layer(ctx: "void*", deposition_layer_thickness: "void*",
-                                             n_control_volumes: "int") -> "int":
+def update_internal_deposit_layer(
+    ctx: "void*", deposition_layer_thickness: "void*", n_control_volumes: "int"
+) -> "int":
     """
-    **c++ signature** : ``HOOK_UPDATE_PLUGINS_INTERNAL_DEPOSIT_LAYER(void* ctx, void* deposition_layer_thickness,
+    **c++ signature** : ``HOOK_UPDATE_INTERNAL_DEPOSIT_LAYER(void* ctx, void* deposition_layer_thickness,
     int n_control_volumes)``
 
     Internal simulator hook to evaluate the thickness of the deposited layer at the inside of the pipeline walls.
@@ -1356,20 +1357,10 @@ def update_plugins_internal_deposit_layer(ctx: "void*", deposition_layer_thickne
         :linenos:
         :emphasize-lines: 1
 
-        HOOK_UPDATE_PLUGINS_INTERNAL_DEPOSIT_LAYER(
+        HOOK_UPDATE_INTERNAL_DEPOSIT_LAYER(
             ctx, deposition_layer_thickness, n_control_volumes)
         {
-            errcode = alfasim_sdk_api.get_field_id(
-                ctx, &liquid_id, "liquid");
-            // Convertion from void* to double* and getting the
-            // array range related to liquid field
-            double* liquid_mass_source =
-                (double*) mass_source + n_control_volumes * liq_id;
-            // Make some calculations and add it to liquid_mass_source.
-            // In this example, we add a mass source of 3.1415 kg/s to all control volumes.
-            for (int i = 0; i < n_control_volumes; ++i) {
-                liquid_mass_source[i] = 3.1415; // [kg/s]
-            }
+
             return OK;
     }
 
@@ -1416,6 +1407,6 @@ specs = HookSpecs(
         env_temperature,
         calculate_entrained_liquid_fraction,
         # Internal Deposition Layer
-        update_plugins_internal_deposit_layer,
+        update_internal_deposit_layer,
     ],
 )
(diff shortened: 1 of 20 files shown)


alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,52 @@ def calculate_entrained_liquid_fraction(
     """
 
 
+def update_plugins_internal_deposit_layer(ctx: "void*", deposition_layer_thickness: "void*",
```
nicoddemus: I suggest 

```suggestion
def update_internal_deposit_layer(ctx: "void*", deposition_layer_thickness: "void*",
```

The fact that we are already in a plugin is enough. I see we have `update_plugins_secondary_variables` and you probably followed suit, but I think my suggestion is better, just not worth updating `update_plugins_secondary_variables` too.

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,52 @@ def calculate_entrained_liquid_fraction(
...
+    The plugin is supposed to change the given ``deposition_layer_thickness`` array pointer. Its values are contiguous
+    in memory and the dimension is given by ``n_control_volumes``. It has unit equal to ``[m]``.
+
+    :param ctx: ALFAsim's plugins context
+    :param deposition_layer_thickness: Thickness of the internal deposition layer
+    :param n_control_volumes: Number of control volumes
+    :returns: Return OK if successful or anything different if failed
+
+    Example of usage:
+    TODO!!!!!!!!
```
nicoddemus: Better remove the entire block instead of leaving the wrong example in place... this way it is easier to miss the TODO and end up with the wrong example.

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,52 @@ def calculate_entrained_liquid_fraction(
...
+def update_plugins_internal_deposit_layer(ctx: "void*", deposition_layer_thickness: "void*",
+                                             n_control_volumes: "int") -> "int":
+    """
+    **c++ signature** : ``HOOK_UPDATE_PLUGINS_INTERNAL_DEPOSIT_LAYER(void* ctx, void* deposition_layer_thickness,
+    int n_control_volumes)``
+
+    Internal simulator hook to evaluate the thickness of the deposited layer at the inside of the pipeline walls.
+    This is called at the beginning of accounting the diameter reduction.
+
+    The plugin is supposed to change the given ``deposition_layer_thickness`` array pointer. Its values are contiguous
```
nicoddemus: Hmmm is not entirely clear if the plugin should add/reduce from the existing values, or set a value blindly. Also will does play out when multiple plugins implement this hook?

If multiple plugins should just "add" their thickness to each control volume, won't this be a problem in case the plugin needs to track the amount of deposited layer from the previous time step? Is the plugin supposed to track this using their own variables, initialized in `initialize`?

williamjamir:
The linting is failing, you probably don't have pre-commit set up =/

From the alfasim-sdk main folder execute.
```python
pre-commit install
```
This way, every new commit will execute the linting (similar do `ff`).

In order to execute the pre-commit and fix the linting from this PR execute:
```python
pre-commit run --all-files --show-diff-on-failure
```

alfasim_sdk/hook_specs.py
sofiapbittencourt: Done it

alfasim_sdk/hook_specs.py
sofiapbittencourt: The plugin should set the value calculated for the thickness at the time-step and not set the difference of thickness between two time-steps. 
So, as an example, the wax plugin calculates the first wax deposit layer thickness, saves this value as a plugin secondary variable (called `wax_thickness`),  calculates the variation of the thickness between 2 time-steps and it evaluates the new thickness by summing the difference with `wax_thickness_old`.

sofiapbittencourt: About multiple plugins are not yet supported for all hooks, but the way is done is analogous to other hooks (such as `calculate_mass_source_term`) and it is ready for when the task for multiple plugins be done.

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,43 @@ def calculate_entrained_liquid_fraction(
...
+        HOOK_UPDATE_INTERNAL_DEPOSIT_LAYER(
+            ctx, deposition_layer_thickness, n_control_volumes)
+        {
+
+            return OK;
+    }
+
+    In the example above is shown how to manage the ``mass_source`` array to get the mass source term array related to a
+    specific field (`liquid field` in this case). Note that ``liquid_mass_source`` has size equal to ``n_control_volumes``.
+
```
williamjamir: I think that this paragraph came from a copy/paste  😋 
- [ ] Update paragraph

alfasim_sdk/hook_specs.py
ggrbill: I think it would be nice to add an example. All documentation has an example. Take a look at them to have some ideas.

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,43 @@ def calculate_entrained_liquid_fraction(
...
+    The plugin is supposed to change the given ``deposition_layer_thickness`` array pointer. Its values are contiguous
+    in memory and the dimension is given by ``n_control_volumes``. It has unit equal to ``[m]``.
+
+    :param ctx: ALFAsim's plugins context
+    :param deposition_layer_thickness: Thickness of the internal deposition layer
+    :param n_control_volumes: Number of control volumes
+    :returns: Return OK if successful or anything different if failed
+
+    Example of usage:
+    TODO!!!!!!!!
```
nicoddemus:
```suggestion

```

alfasim_sdk/hook_specs.py
nicoddemus: Agreed. 👍

sofiapbittencourt: - [ ] Fix example of use

docs/source/06_solver_hooks.rst
```diff
@@ -154,6 +154,17 @@ implementing the following Solver `Hooks`.Otherwise, the particle size will be c
...
 .. autofunction:: alfasim_sdk.hook_specs.update_particle_diameter_of_solids_fields
 
+Internal deposit layer
+----------------------
+
+When a new phase is added to the hydrodynamic model using the :class:`~alfasim_sdk.types.AddPhase` type, it is possible
+that this phase will deposit inside of the pipeline walls. In this case, the thickness of the deposited layer can be
+calculated through the following Solver `Hook`. If not implemented, the layer thickness will be considered equal
+to zero meters.
+
```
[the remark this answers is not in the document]
sofiapbittencourt: What about:
```suggestion
When a new phase is added to the hydrodynamic model using the :class:`~alfasim_sdk.types.AddPhase` type, it is possible to consider the deposition inside the pipeline walls. If so, calculate the thickness of the deposited layer on a given phase through the usage of :py:func:`~alfasim_sdk.hook_specs.update_internal_deposit_layer`. By default, the layer thickness will be considered equal to zero meters.
```

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,82 @@ def calculate_entrained_liquid_fraction(
...
+                // Sum this variation with the thickness of the older time step
+                for (int i = 0; i < n_control_volumes; ++i) {
+                    deposit_layer_thickness[i] = deposit_layer_thickness_old[i] +  d_deposit_layer_dt * dt; // [m]
+                }
+            }
+
+            return OK;
+    }
+
+    In the example above is shown how to manage the ``deposit_layer_thickness`` array for each control volume.
```
ggrbill: Nice instruction!!!

docs/source/06_solver_hooks.rst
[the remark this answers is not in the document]
williamjamir: > What about:

Even better!

williamjamir:
Please, update the `HISTORY.rst` file ;)

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,82 @@ def calculate_entrained_liquid_fraction(
     """
 
 
+def update_internal_deposit_layer(
```
arthursoprana: For consistency with the nomenclature in ALFAsim, call it `update_internal_deposition_layer`

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,82 @@ def calculate_entrained_liquid_fraction(
...
+            auto* deposit_layer_thickness = (double*) (deposit_layer_thickness_raw_ptr);
+
+            double dt = -1.0;
+            errcode = alfasim.get_simulation_quantity(ctx, &dt, TimestepScope::CURRENT, (char*) "dt");
+            if (errcode != 0) {
+                return errcode;
+            }
+
+            // Handle first time step, because you won't have the previously information
+            if (is_first_time_step){
```
tarcisiofischer: Initialize this variables somewhere.
As it is, the user can't know where this comes from.

alfasim_sdk/hook_specs.py
```diff
@@ -1333,6 +1333,82 @@ def calculate_entrained_liquid_fraction(
...
+            ctx, deposition_layer_thickness, n_control_volumes)
+        {
+            auto errcode = -1;
+            void* deposit_layer_thickness_raw_ptr = nullptr;
+            errcode = alfasim.get_plugin_variable(
+                ctx, &deposit_layer_thickness_raw_ptr, "deposit_layer_thickness", 0, TimestepScope::CURRENT, &size);
+            if (errcode != 0) {
+                return errcode;
+            }
+            auto* deposit_layer_thickness = (double*) (deposit_layer_thickness_raw_ptr);
```
tarcisiofischer: Can you explain why this code block is necessary, since `deposition_layer_thickness` is being passed as parameter?
What is the difference between `deposition_layer_thickness` and `deposit_layer_thickness`?

alfasim_sdk/hook_specs.py
```diff
@@ -1369,5 +1445,7 @@ def calculate_entrained_liquid_fraction(
         friction_factor,
         env_temperature,
         calculate_entrained_liquid_fraction,
+        # Internal Deposition Layer
```
tarcisiofischer: Remove the comment, it doesn't seem to add any useful information here

docs/source/06_solver_hooks.rst
```diff
@@ -154,6 +154,17 @@ implementing the following Solver `Hooks`.Otherwise, the particle size will be c
 
 .. autofunction:: alfasim_sdk.hook_specs.update_particle_diameter_of_solids_fields
 
+Internal deposit layer
+----------------------
+
+When a new phase is added to the hydrodynamic model using the :class:`~alfasim_sdk.types.AddPhase` type, it is possible
```
tarcisiofischer: Why the docs are duplicated, but sightly different, in those two paragraphs?

alfasim_sdk/hook_specs.py
tarcisiofischer: Do it everywhere (function names, variable names and docs)

sofiapbittencourt: - [ ] change name

alfasim_sdk/hook_specs.py
sofiapbittencourt: - [ ] Initialize this variables

alfasim_sdk/hook_specs.py
sofiapbittencourt: I'm sorry `deposition_layer_thickness` and `deposit_layer_thickness` are the same, just with a typo and I don't need this block, this was probably a copy and paste

sofiapbittencourt: - [ ] remove get_plugin_variable for deposit_layer_thickness

alfasim_sdk/hook_specs.py
sofiapbittencourt: - [ ] Remove the comment

docs/source/06_solver_hooks.rst
sofiapbittencourt: I don't know, maybe it was a merge or a pull with conflict that I didn't see it. I will delete one of them

sofiapbittencourt: - [ ] Remove duplicate paragraph

PR #91, opened 2020-03-17
Add new api functions to the UCM scope

ggrbill:
These two new functions are in the new UCM Scope.

This is the First PR and it is just a preparation. The Docs were not linked yet because the functions may change in the next PR.

### DIFF
--- a/alfasim_sdk_api/api.h
+++ b/alfasim_sdk_api/api.h
@@ -595,7 +595,7 @@ DLL_EXPORT int set_wall_layer_property(void* ctx, int control_volume, int wall_l
 /*!
     Gets the current UCM (unit cell model) input data for friction factor calculation.
     Any available variable by this function is considered for a unit cell, which means
-    that there are variables with one value and there are varailbes with two values
+    that there are variables with one value and there are variables with two values
     related to the two phase system (GAS and LIQUID).
     If a given variable name is not available, a #NOT_AVAILABLE_DATA error is returned.
 
@@ -616,10 +616,10 @@ DLL_EXPORT int set_wall_layer_property(void* ctx, int control_volume, int wall_l
     - `"sigma"`: Gas-liquid Surface Tension [N/m]
 
     @param[in] ctx ALFAsim's plugins context.
-    @param[out] out Variable values array.
+    @param[out] out Variable value.
     @param[in] var_name String with the variable name. See the list of possible values above.
-    @param[in] phase_id Phase Id, When it is Gas phase the Id must be 0
-                          when it is Liquid phase the Id must be 1.
+    @param[in] phase_id A #TwoPhaseSystem value. When the requested variable is not associated
+                            to a phase any value can be passed.
     @return An #error_code value.
 */
 DLL_EXPORT int get_ucm_friction_factor_input_variable(
@@ -639,8 +639,8 @@ DLL_EXPORT int get_ucm_friction_factor_input_variable(
     @param[in] ctx ALFAsim's plugins context.
     @param[out] S_w Wetted Perimeters [m].
     @param[out] S_i Interface Perimeter [m].
-    @param[out] H   Phase height [m]. For annular flow, H[0] is the core diameter and H[1] is two
-                     times the liquid film.
+    @param[out] H   Phase height [m]. For annular flow, H[GAS] is the core diameter and H[LIQUID] is
+                     the total liquid film height.
     @param[in] alpha_G  Unit Cell Gas Volume Fraction [m3 of `gas phase` /m3 of mixture].
     @param[in] D        Unit Cell Inner Diameter [m].
     @param[in] phase_id Phase Id, When it is Gas phase the Id must be 0
@@ -652,8 +652,8 @@ DLL_EXPORT int get_ucm_fluid_geometrical_properties(
     double** S_w,
     double* S_i,
     double** H,
-    const double alpha_G,
-    const double D,
+    double alpha_G,
+    double D,
     int phase_id
 );
 
--- a/alfasim_sdk_api/api.h
+++ b/alfasim_sdk_api/api.h
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
 */
 DLL_EXPORT int set_wall_layer_property(void* ctx, int control_volume, int wall_layer_id, enum WallLayerProperty property_id, double new_value);
 
+/*!
+    Gets the current UCM (unit cell model) input data for friction factor calculation.
+    Any available variable by this function is considered for a unit cell, which means
+    that there are variables with one value and there are varailbes with two values
+    related to the two phase system (GAS and LIQUID).
+    If a given variable name is not available, a #NOT_AVAILABLE_DATA error is returned.
+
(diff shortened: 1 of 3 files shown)


alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
 */
 DLL_EXPORT int set_wall_layer_property(void* ctx, int control_volume, int wall_layer_id, enum WallLayerProperty property_id, double new_value);
 
+/*!
+    Gets the current UCM (unit cell model) input data for friction factor calculation.
+    Any available variable by this function is considered for a unit cell, which means
+    that there are variables with one value and there are varailbes with two values
```
tarcisiofischer: typo varailbes

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+    - `"rho"`: Density [kg/m3]
+    - `"mu"`: Viscosity [Pa.s]
+    - `"U"`: Velocity [m/s]
+
+    It is important to know that the listed `variable_name`s are not available in any phase, only for
+    two phase systems, in which `Gas id` is 0 (zero) and `Liquid id` (sum of all liquid phases) is 1 (One).
+    Because of that, the #error_code must be checked.
+
+    List of `variable_name` with one value:
+    - `"D"`: Unit Cell Inner Diameter [m]
```
tarcisiofischer: Missing many vars:

```
    EXPORT_VARIABLE("alpha", context->alpha[phase_id]);
    EXPORT_VARIABLE("rho", context->rho[phase_id]);
    EXPORT_VARIABLE("mu", context->mu[phase_id]);
    EXPORT_VARIABLE("U", context->u[phase_id]);
    EXPORT_VARIABLE("D", context->D);
    EXPORT_VARIABLE("ks", context->ks);
    EXPORT_VARIABLE("theta", context->theta);
    EXPORT_VARIABLE("sigma", context->sigma);
```

It'll be a pain to maintain the two synchronized, though

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+    - `"ks"`: Roughness [m]
+    - `"theta"`: Inclination of the Unit Cell [rad]
+    - `"sigma"`: Gas-liquid Surface Tension [N/m]
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] out Variable values array.
```
tarcisiofischer: Isn't the output a single value?

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+    @param[in] var_name String with the variable name. See the list of possible values above.
+    @param[in] phase_id Phase Id, When it is Gas phase the Id must be 0
```
tarcisiofischer: Isn't there an enum for phase ids, instead of those raw integers?

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+    During the implementation of any HOOK related to the UCM friction factor, this
+    function provides the following fluid geometrical properties:
+    - `"S_w"`: Wetted perimeters of phases [m].
+    - `"S_i"`: Interface perimeter [m].
+    - `"H"`: Phase height [m].
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] S_w Wetted Perimeters [m].
+    @param[out] S_i Interface Perimeter [m].
+    @param[out] H   Phase height [m]. For annular flow, H[0] is the core diameter and H[1] is two
```
tarcisiofischer: Is it common to be "two times the liquid film"? Could we divide by two and pass the liquid film instead? Does it make any sense?

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+                     times the liquid film.
+    @param[in] alpha_G  Unit Cell Gas Volume Fraction [m3 of `gas phase` /m3 of mixture].
+    @param[in] D        Unit Cell Inner Diameter [m].
+    @param[in] phase_id Phase Id, When it is Gas phase the Id must be 0
```
tarcisiofischer: Isn't there an enum for phase ids, instead of those raw integers?

alfasim_sdk_api/api.h
```diff
@@ -592,4 +592,69 @@ DLL_EXPORT int get_wall_layer_id(void* ctx, int control_volume, const char* mate
...
+                          when it is Liquid phase the Id must be 1.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_ucm_fluid_geometrical_properties(
+    void* ctx,
+    double** S_w,
+    double* S_i,
+    double** H,
+    const double alpha_G,
```
tarcisiofischer: lose the `const`... They are copy values anyway

alfasim_sdk_api/api.h
ggrbill: No, it is not missing any variable. Please, take a look at lines 603-606.
There are two list of vars.

alfasim_sdk_api/api.h
ggrbill: Please, take a look at lines 603-606.

alfasim_sdk_api/api.h
ggrbill: - [ ] Create new Enum called TwoPhase.

alfasim_sdk_api/api.h
ggrbill: - [ ] Remove `const`

alfasim_sdk_api/api.h
ggrbill: - [ ] Typo

alfasim_sdk_api/api.h
ggrbill: I will create it.

alfasim_sdk_api/api.h
ggrbill: I think it is not the usual way to deal with annular flows in Unit Cell Model. 
Do you agree @arthursoprana ?

arthursoprana: When I first read the comment I was thinking that the ideal would be `H` to be the liquid film. But then the definitions would change for stratified and annular geometries. In the literature the liquid film thickness is usually represented by `δ`. I currently don't know what would be better and probably if I implement any friction factor model in the future I will always need to check the definition of `H` since it is not the film thickness anyway.

If we change, for the annular geometry `H` to be the actual film thickness, we'll need to review the friction factor equations in order to make sure they  are not already multiplying by `0.5`.

If you don't want to change that, perhaps call it `Total phase height`.

ggrbill: - [ ] `Total phase height`

PR #95, opened 2020-04-08
Fb aup 61 support 9 fields

edersongrein:
Add two new mfd models for three-phase flow

### DIFF
--- a/alfasim_sdk/constants.py
+++ b/alfasim_sdk/constants.py
@@ -1,6 +1,5 @@
 from enum import Enum
 
-
 GAS_PHASE = "gas"
 """
 Constant to identify the gas phase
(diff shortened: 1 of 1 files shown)


edersongrein:
The commits are a mess indeed. Let me fix it

alfasim_sdk/constants.py
```diff
@@ -75,6 +75,12 @@ class HydrodynamicModelType(Enum):
     ThreeLayersWaterWithCO2 = (
         "hydrodynamic_model_3_layers_water_with_co2"  # Under Development
     )
+    ThreeLayersSevenFieldsGasOilWater = (
+        "hydrodynamic_model_3_layers_7_fields_gas_oil_water"
+    )
+    ThreeLayersNineFieldsGasOilWater = (
+        "hydrodynamic_model_3_layers_9_fields_gas_oil_water"
+    )
```
williamjamir, line 83: Please update the docstring as well from this method, since it will be used automatically for documentation.

edersongrein: Of course. I was forgetting

PR #103, opened 2020-10-02
Fix Deploy action from ALFAsim-SDK CI

williamjamir:


.github/workflows/main.yml
```diff
@@ -78,8 +78,8 @@ jobs:
     - name: Build Package
       run: |
         python -m pip install --upgrade pip
-        curl -fsS -o get-poetry.py https://raw.githubusercontent.com/sdispater/poetry/master/get-poetry.py
-        python get-poetry.py --preview -y
+        curl -fsS -o get-poetry.py https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py
+        python get-poetry.py --version 1.1.0
```
nicoddemus, line 82: Can you leave a comment on why this version? Also should it pin exactly, or can we use something like `>=1.1`?

williamjamir: Yep, I will add a comment.

I think that we can keep as `>=1.1`, the version 1.1.0 in specific is due to fix for packages being installed directly from git.

williamjamir: Hmmm thinking again about this issue, 
Do remember why we are installing hookman from git and not PyPi?

I will try to install the python-hookman directly from PyPI and see if everything works smoothly

nicoddemus: I don't remember, but I agree PyPI is preferable. The builds have failed, but due to a test failure.

PR #104, opened 2020-10-02
Update docs build dependencies

williamjamir:
I believe that everything is working normally now.
https://alfasim-sdk.readthedocs.io/en/update-docs-build-dependencies/

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,6 +62,7 @@ jobs:
           tox -e linting
 
   deploy:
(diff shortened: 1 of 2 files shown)


prusse-martin (approved):
I will approve.
But ask one thing.
Should the deploy be running on prs?
@nicoddemus

nicoddemus:
Ideally not, specially if they will fail. 

Here's how we configure it in pytest so it only runs the deploy job on tags:

https://github.com/pytest-dev/pytest/blob/cb0a13a523e7d9f1e9f3bd77095a9f58adb4e7b7/.github/workflows/main.yml#L186-L187

PR #105, opened 2020-10-05
Update signature of hook update_internal_deposition_layer

arthursoprana:


williamjamir:
Please update [Histoy](https://github.com/ESSS/alfasim-sdk/blob/master/HISTORY.rst) file ;)

alfasim_sdk/hook_specs.py
```diff
@@ -1450,7 +1453,7 @@ def update_internal_deposition_layer(
         :emphasize-lines: 1
 
         HOOK_UPDATE_INTERNAL_DEPOSITION_LAYER(
-            ctx, deposition_layer_thickness, n_control_volumes)
+            ctx, thickness, density, heat_capacity, thermal_conductivity, n_control_volumes)
```
ggrbill, line 1456: It would be nice to add an example of setting the deposition layer properties since it will be set zero by default.

Perhaps add and warning in the documentation saying that if the properties are not set a zero value is assumed as default.

ggrbill (approved):

PR #106, opened 2020-10-22
Update tox.ini

williamjamir:


tox.ini
```diff
@@ -12,6 +12,7 @@ deps=
     poetry
 
 commands =
+    python -m pip install -U pip
```
nicoddemus, line 15: You should use `download=true` instead to update pip:

https://github.com/pytest-dev/pytest/blob/e8504e04f314528bc02eba13d76048295812fac2/tox.ini#L118-L119

nicoddemus (changes requested):

williamjamir:
Closing this PR in favor of #107

PR #112, opened 2020-11-09
Export only classes and functions that can be used by users on alfasim-sdk module

williamjamir:
ASIM-3940


HISTORY.rst
```diff
@@ -2,6 +2,12 @@
 History
 =======
 
+0.7.0 (Unreleased)
+==================
+
+* Released with ALFAsim 1.8.0.
+* Add support for WellConectionPort on CaseDescription
```
oliveira-mauricio, line 9: Where did this come from? I believe the support is there since 0.6.0.

williamjamir: Nops, the release was first made with the content on ALFAsim.
After that, the modification of WellConnectionPort was introduced.

image

oliveira-mauricio: Hmm nevermind then. The feature itself was done earlier (merged to master 2 days before the tag 0.6.0 was created), but since you were already working on the migration it unsynced, but no big deal.

oliveira-mauricio (approved):

prusse-martin (approved):

src/_alfasim_sdk/version.py
```diff
@@ -0,0 +1 @@
+__version__ = "0.7.0"
```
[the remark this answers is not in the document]
williamjamir: This file was moved from `alfasim_sdk\_version.py` to `_alfasim_sd\version.py`.
Notice the file name changed from _version to version.py

PR #117, opened 2020-11-18
Add method ensure_unique_name

williamjamir:
ASIM-3987

### DIFF
--- a/src/_alfasim_sdk/alfacase/case_description.py
+++ b/src/_alfasim_sdk/alfacase/case_description.py
@@ -2031,14 +2031,29 @@ class CaseDescription:
 
         def get_error_msg():
             output = []
-            for key, value in duplicate_names.items():
-                output.append(f"{key}:")
-                formatted_names = "\n    - ".join(name for name in value)
-                output.append(f"    - {formatted_names}")
+            for key, value in sorted(duplicate_names.items()):
+                if value:
+                    output.append(f"{key}:")
+                    formatted_names = "\n    - ".join(name for name in value)
+                    output.append(f"    - {formatted_names}")
             return "\n".join(output)
 
-        if any(value for key, value in sorted(duplicate_names.items())):
+        if any(value for key, value in duplicate_names.items()):
             raise InvalidReferenceError(
                 f"Elements that can be referenced must have a unique name, found multiples definitions of the following items:\n"
+                f"{get_error_msg()}"
+            )
+
+        # Check unique name between elements
+        duplicate_names["Nodes and Wells"] = get_duplicate_keys(
+            Counter(all_wells_names + all_node_names)
+        )
+        duplicate_names["Pipes and Wells"] = get_duplicate_keys(
+            Counter(all_wells_names + all_pipes_names)
+        )
+
+        if any(value for key, value in duplicate_names.items()):
+            raise InvalidReferenceError(
+                f"Some different type of elements needs to have unique name between them, found duplicated names for the following items:\n"
                 f"{get_error_msg()}"
             )
--- a/src/_alfasim_sdk/alfacase/case_description.py
+++ b/src/_alfasim_sdk/alfacase/case_description.py
@@ -1983,7 +1983,7 @@ class CaseDescription:
         self._check_pvt_model_references()
         self._check_restart_file()
         self._check_fluid_references()
-        self.ensure_unique_name()
+        self.ensure_unique_names()
 
     def reset_invalid_references(self):
         """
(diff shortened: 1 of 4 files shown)


src/_alfasim_sdk/alfacase/case_description.py
```diff
@@ -1991,3 +1992,52 @@ def reset_invalid_references(self):
         self._check_pvt_model_files(reset_invalid_reference=True)
         self._check_pvt_model_references(reset_invalid_reference=True)
         self._check_fluid_references(reset_invalid_reference=True)
+
+    def ensure_unique_name(self):
```
nicoddemus, line 1996:
```suggestion
    def ensure_unique_names(self):
```

src/_alfasim_sdk/alfacase/case_description.py
```diff
@@ -1991,3 +1992,52 @@ def reset_invalid_references(self):
...
+        """
+        Ensure that elements that can be referenced by name have a unique name.
```
nicoddemus, line 1998:
```suggestion
        Ensure that elements that can be referenced by name have a unique name, 
        raising `InvalidReferenceError` if some elements have the same name .
```

src/_alfasim_sdk/alfacase/case_description.py
```diff
@@ -1991,3 +1992,52 @@ def reset_invalid_references(self):
...
+
+        def get_error_msg():
+            output = []
+            for key, value in duplicate_names.items():
+                output.append(f"{key}:")
+                formatted_names = "\n    - ".join(name for name in value)
+                output.append(f"    - {formatted_names}")
+            return "\n".join(output)
+
+        if any(value for key, value in duplicate_names.items()):
```
nicoddemus, line 2039:
```suggestion
        if any(value for key, value in sorted(duplicate_names.items())):
```

nicoddemus (approved):
Left a few minor comments.

prusse-martin (approved):

src/_alfasim_sdk/alfacase/case_description.py
williamjamir: Sorry I didn't understand what the benefit of sorting the list here, I'm just checking if any of the entries have a list with elements
```python
>>>  d1 = {'a': []}
>>> d2 = {'a': [], 'b': [1]}
>>> any(value for key, value in (d1.items())
False
>>> any(value for key, value in d2.items())
True
```

tests/alfacase/test_case_description.py
```diff
@@ -1054,3 +1054,103 @@ def test_invalid_fluid_reference_on_wells(default_well):
...
+        Materials:
+            - Material
+        PVT:
+            - PVT1
+        Fluids:
+            - Fluid 0
+            - Fluid 1"""
+    )
+    with pytest.raises(InvalidReferenceError, match=re.escape(expected_msg)):
+        case.ensure_unique_name()
```
oliveira-mauricio, line 1156: We also don't want:
- Pipes and Wells with same names; (because of OutputDefinition)
- Nodes and Wells with same names; (because of Edge source/target)

We should address those cases too (there may be others, but I can't remember other ones right now).

In the future, once we have _Equipment Trend_ in case description, edge's equipment will need to be unique accross the entire network, but since it is not available yet, you can leave that one out.

src/_alfasim_sdk/alfacase/case_description.py
nicoddemus: Oh you are right, I meant sorting the dict in `get_error_msg`. Updating.

src/_alfasim_sdk/alfacase/case_description.py
```diff
@@ -1991,3 +1992,52 @@ def reset_invalid_references(self):
...
+        duplicate_names["Pipes"] = get_duplicate_keys(Counter(all_pipes_names))
+        duplicate_names["Wells"] = get_duplicate_keys(Counter(all_wells_names))
+        duplicate_names["Walls"] = get_duplicate_keys(Counter(all_walls_names))
+        duplicate_names["Materials"] = get_duplicate_keys(Counter(all_material_names))
+        duplicate_names["PVT"] = get_duplicate_keys(Counter(all_pvt_names))
+        duplicate_names["Fluids"] = get_duplicate_keys(Counter(all_fluids))
+
+        def get_error_msg():
+            output = []
+            for key, value in duplicate_names.items():
```
nicoddemus, line 2033:
```suggestion
            for key, value in sorted(duplicate_names.items()):
```

oliveira-mauricio (approved):

PR #119, opened 2020-11-19
Fix ALFASim-SDK docs to work with src directory

williamjamir:


docs/source/02_alfasim_sdk_by_example.rst
```diff
@@ -216,7 +216,7 @@ needed to load and unload the |sdk| API, in which will allows the plugin to use
     }
 
 Then, since the plugin wants to calculate its own secondary variable, the
-:func:`HOOK_UPDATE_PLUGINS_SECONDARY_VARIABLES <alfasim_sdk.hook_specs.update_plugins_secondary_variables>` must be implemented.
+:func:`HOOK_UPDATE_PLUGINS_SECONDARY_VARIABLES <_alfasim_sdk.hook_specs.update_plugins_secondary_variables>` must be implemented.
```
nicoddemus, line 219: This link should use the public interface, rather than reference an internal module...

williamjamir: `hook_specs` is not intended to be used by ALFASim-SDK users and for this reason I kept as private, 

Basically `hook_specs` is an input for an internal tool (python-hookman) to generate the header files (hooks_specs.h) for the end-users.

docs/source/99_2_user_interface_reference.rst
```diff
@@ -77,5 +77,5 @@ Status
 Context
 -------
 
-.. automodule:: alfasim_sdk.context
+.. automodule:: _alfasim_sdk.context
```
nicoddemus, line 80: Why are `status` and `context` not available in the public API?

williamjamir: *Status*:  The status module has only two classes, that are publicly available, I could argue that this was only  "laziness" to documented all classes automatically by using the `automodule`. I will add these, two classes, explicitly here by accessing them thought the public API

*Context*: The context module just has classes that the user will interact with, but will not have to instantiate, for example:

We have the class `PluginInfo`, `PipelineInfo`, `NodeInfo` that are accessed through the Context when executing `enable-expression`, `visible-expression` or `status monitor`
```python
@alfasim_sdk.hookimpl
def alfasim_get_status(ctx):
    ....
```

And by accessing the methods the uses has info about the current state of the application (read-only values)
```python
>>>  ctx.get_pipelines()[0]
PipelineInfo(name='Pipe 1 > Pipeline', [ ... ]) 

>>> ctx.get_nodes()
[NodeInfo(name='Node 1', number_of_phases_from_associated_pvt=2)] 
``` 

So the context-related classes should be on the private API, right?
And for these cases, can I stick with the  `automodule`  approach?

docs/source/02_alfasim_sdk_by_example.rst
```diff
@@ -146,8 +146,8 @@ make use of the :ref:`ALFAsim-SDK C/C++ API <sdk_api>` in order to fetch informa
 At this point, we are going to implement the :ref:`solver_hooks` that updates the secondary variable declared from
 :file:`myplugin.py` file and retrieve the ``Liquid Velocity`` from the |alfasim|'s Solver.
 
-First, we need to implement two mandatory hooks, the :py:func:`HOOK_INITIALIZE <alfasim_sdk.hook_specs.initialize>` and
-the :py:func:`HOOK_FINALIZE <alfasim_sdk.hook_specs.finalize>`
+First, we need to implement two mandatory hooks, the :py:func:`HOOK_INITIALIZE <__alfasim_sdk.hook_specs.initialize>` and
```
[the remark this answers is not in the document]
williamjamir: This is a wrong reference, 2d5e38c fix this issue.

docs/source/99_2_user_interface_reference.rst
nicoddemus: I see, thanks for the explanation!

Let's document `status` using the public links, and `context` can keep the private links. 👍

nicoddemus (approved):

nicoddemus:
Great job!

PR #123, opened 2020-11-30
Add API reference section for ALFAcase

williamjamir:
Link for the demo: https://alfasim-sdk.readthedocs.io/en/fb-asim-3584-doc-for-alfacase/alfacase/index.html


nicoddemus (approved):

docs/source/conf.py
```diff
@@ -77,6 +83,11 @@
 .. |marker_2| image:: /_static/images/marker_2.png
     :scale: 80%
 
+.. # define a hard space useful for literal-blocks that has empty content.
+.. |space| raw:: html
```
prusse-martin, line 87: I need to check how this affects to final layout...

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
 breathe==4.22.1
```
prusse-martin, line 1: Not related to this file.
(I did not find a way to add a comment not related to a files in an open review)

Did you notice that when a tab is changed ALL tab components are affected and the current position in the document is lost?

It affect all tabs with the same label in the page.
```
function onLabelClick() {
  // Activate other labels with the same text.
  for (label of labels_by_text[this.textContent]) {
    if (label === this) continue;
    label.previousSibling.checked = true;
  }
}
```

docs/requirements.txt
```diff
@@ -1,3 +1,4 @@
...
 sphinx-click==2.5.0
 pydata-sphinx-theme
+sphinx-inline-tabs
```
prusse-martin, line 4: `sphinx-tabs` are not grouped by default, the author must explicitly declare them as grouped:

https://github.com/executablebooks/sphinx-tabs

`sphinx-inline-tabs` have this grouping behavior as a feature.

> Synchronisation: Tabs with the same label all switch with a single click.

prusse-martin (changes requested):

docs/source/conf.py
prusse-martin: I did not notice any significant blank space.
Maybe I was looking at the wrong place.

oliveira-mauricio (approved):

docs/source/alfacase/api_reference.rst
```diff
@@ -2,3 +2,214 @@ Full API Reference
...
+
+.. autoclass:: alfasim_sdk.TimeOptionsDescription()
+
+.. autoclass:: alfasim_sdk.NumericalOptionsDescription()
+
+
+Outputs
+-------
+
+.. autoclass:: alfasim_sdk.CaseOutputDescription()
```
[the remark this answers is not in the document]
williamjamir: Because I want to hide the class signature, which is confusing in my opinion.

image

docs/source/alfacase_definitions/NumericalOptionsDescription.txt
```diff
@@ -0,0 +1,42 @@
+.. rubric:: Definitions
+
+.. tab:: CaseDescription
+
+    .. parsed-literal::
+
+        class NumericalOptionsDescription
```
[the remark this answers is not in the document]
williamjamir: This section only shows the class name and their fields with a cross-referencing link to the definition.
I don't know (and couldn't find as well) another solution to this "requirement" which lists only the attributes from a class

docs/requirements.txt
williamjamir: >Did you notice that when a tab is changed ALL tab components are affected

Yes, this is actually a feature all tabs with the same name are syncronized.

> and the current position in the document is lost?

This is an, unfortunately, a side effect  =/ 
Since different tabs can have different outputs that affect the page height, this difference is more noticeable at the bottom.

docs/requirements.txt
williamjamir: Sorry I didn't get your point.
Do you want to disable the synchronization?

prusse-martin: I just think that content swinging to top and bottom are a bit distracting.

prusse-martin (approved):

PR #124, opened 2020-12-02
Add docs for ALFAcase file

williamjamir:
Link for the online docs
https://alfasim-sdk.readthedocs.io/en/fb-asim-3584-add-introduction-for-alfacase/


docs/source/alfacase/02_alfacase_syntax.rst
```diff
@@ -0,0 +1,216 @@
+.. _alfacase-syntax:
+
+ALFAcase Syntax
+===============
+
+
+An :program:`ALFAcase` file is a text-based file that must be written following the YAML syntax, only a restricted subset of the YAML specification
+is followed and the following section and this section covers the support YAML syntax how to configure a project.
```
nicoddemus, line 8:
```suggestion
is supported and the this section and the next cover the differences.
```

But I'm not sure we should add these here: while I appreciate the intention, I think a link to the Strict YAML docs is more effective and can replace this entire section.

docs/source/alfacase/03_alfacase_example.rst
```diff
@@ -0,0 +1,121 @@
...
+    numerical_options: numerical_options_description_schema⠀
+    ipr_models: ipr_models_description_schema⠀
+    pvt_models: pvt_models_description_schema⠀
+    tracers: tracers_description_schema⠀
+    outputs: case_output_description_schema⠀
+
+    ....
+
+We are going to create a simple case with a PVT model, two nodes and one pipe displaying their output.
+Each section can be inspect over the :ref:`alfacase-reference-section`
```
nicoddemus, line 26: Please don't forget to add `:` before blocks, looks bad without it.

docs/source/alfacase/03_alfacase_example.rst
```diff
@@ -0,0 +1,121 @@
+.. _alfacase-example:
+
+ALFAcase Example
+================
+
+In this section, we will create a project from scratch with a simple network and insert it on the ALFAsim application.
```
nicoddemus, line 6: I don't think this page (the file) adds much: we should guide the user to produce a case file using the application at least as a starting point, and then build from there using the Case documentation as guide.

nicoddemus (changes requested):
Overall good work, but I think in this case "less is more": let's not duplicate what can be found elsewhere (and in more details) elsewhere, let's link to it instead.

docs/source/alfacase/01_alfacase_quick_start.rst
```diff
@@ -0,0 +1,43 @@
+.. _alfacase-quick-start-section:
+
+ALFAcase Quick Start
+====================
+
+In this section, it is shown how to create an :program:`ALFAcase` from an existent project, with this exported project
```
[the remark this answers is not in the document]
oliveira-mauricio: Maybe also add somewhere at the index that _ALFACase_ is referenced as _ALFAsim case file_ (like in the app import/export buttons), just to make a link between two terms.

docs/source/alfacase/02_alfacase_syntax.rst
```diff
@@ -0,0 +1,216 @@
...
+
+Bool
+----
+
+``bool`` is case-insensitive and accepts the following options:
+
+    .. code-block:: yaml
+        :caption: boolean syntax
+
+        enable_solver_caching: True     # True
```
oliveira-mauricio, line 208: I would just group up True and False values together instead of mixing them.

docs/source/alfacase/03_alfacase_example.rst
oliveira-mauricio, line 26:
```suggestion
Each section can be inspected over the :ref:`alfacase-reference-section`
```

docs/source/alfacase/03_alfacase_example.rst
```diff
@@ -0,0 +1,121 @@
...
+    pvt_models:
+    default_model: 'Pvt1'
+    tables:
+      'Pvt1': my_pvt_file.tab
+
+    numerical_options:
+    tolerance: 0.0001
+
+    time_options:
+    minimum_timestep:
```
oliveira-mauricio, line 44: These should be indented, right?

docs/source/alfacase/02_alfacase_syntax.rst
williamjamir: I don't know if a link to the StrictYAML would be good for our users since the StrictYAML docs are focused on the DEV using the lib, with mentions and instructions on how to extract and manipulate the data, many details that out of the scope of our user that only needs to fill a file.

I also don't wanna link a cheatsheet with YAML syntax since only a subset of the specification is followed by strictyaml,  and in my view, our users just need to know how to write a "string", "number", "boolean" and a "list with numbers".

docs/source/alfacase/03_alfacase_example.rst
williamjamir: yep, probably I modified by mistake this section after running this snippet through a "prettify code" tool.

docs/source/alfacase/02_alfacase_syntax.rst
williamjamir: Yep, good suggestion =)

docs/source/alfacase/02_alfacase_syntax.rst
nicoddemus: > I don't know if a link to the StrictYAML would be good for our users since the StrictYAML docs are focused on the DEV using the lib

Ahh good point, I misremembered the docs, I thought they had a quick YAML primer but they don't.

> I also don't wanna link a cheatsheet with YAML syntax since only a subset of the specification is followed by strictyaml, and in my view, our users just need to know how to write a "string", "number", "boolean" and a "list with numbers".

I understand, but I also don't want to replicate YAML's documentation here. But OK, I also couldn't find any primer/tutorial which was generic enough, so let's keep it. 👍

oliveira-mauricio:
Not really part of this PR, but looking at the live docs I've noticed this: https://alfasim-sdk.readthedocs.io/en/fb-asim-3584-add-introduction-for-alfacase/alfacase/api_reference.html#equipments

Should be **Equipment**, not _Equipments_

nicoddemus:
> Should be Equipment, not Equipments

While at it, it would be great to search the docs for "equipments" and fixing those. 👍

docs/source/alfacase/03_alfacase_example.rst
williamjamir: > We should guide the user to produce a case file using the application at least as a starting point

Hmm, I don't know if I like this idea, the `ALFAcase` generated from the application has all values filled and it brings too much information for something that should guide the user.

I think that I will split this example into sections, introducing the options and after that instructing how to fill the separately the PVT,  Output, Pipe, and Node entries. 
With this, I think that the user will be ready to follow the API reference and edit a full alfacase generated from the application.

What did you think about this?

prusse-martin (approved):

docs/source/alfacase/03_alfacase_example.rst
nicoddemus: TBH I don't think users will be generating files from scratch, rather they will always build up a small project, generate a case, and edit things from there. But go with your gut feeling then.

nicoddemus:
> While at it, it would be great to search the docs for "equipments" and fixing those

@williamjamir did you try this? There was only one incorrect word?

docs/source/alfacase/03_alfacase_example.rst
oliveira-mauricio: Hmm I'm not sure which will be the common workflow too, but I also feel like creating it from scratch doesn't seem like the way to go (our only user so far is PWPa and they didn't go that path, right?). I can think of three more plausible ways: generating a project and editing it, copying an example or using a template.

But TBH I don't think we should overthink this, it will be a long time before we have real users of this feature, so do as you wish.

docs/source/alfacase/03_alfacase_example.rst
```diff
@@ -0,0 +1,121 @@
...
+
+We are going to create a simple case with a PVT model, two nodes and one pipe displaying their output.
+Each section can be inspected over the :ref:`alfacase-reference-section`:
+
+.. code-block:: yaml
+
+    name: basic_case
+    physics:
+    correlations_package: correlation_package_alfasim
+    hydrodynamic_model: hydrodynamic_model_4_fields
```
oliveira-mauricio, line 33: Shouldn't these be indented too? And also the `pvt_models` section below?

oliveira-mauricio (approved):
Just take a look in .alfacase examples around, some have wrong indentation.

nicoddemus:
The full API looks just great, awesome work!

https://alfasim-sdk--124.org.readthedocs.build/en/124/alfacase/api_reference.html#case

nicoddemus (approved):

docs/source/alfacase/03_alfacase_example.rst
williamjamir: Commit b3e9df5 improves the example section.

docs/source/alfacase/03_alfacase_example.rst
williamjamir: > > While at it, it would be great to search the docs for "equipments" and fixing those
> 
> @williamjamir did you try this? There was only one incorrect word?

Yes, only one occurrence =)

williamjamir:
> > While at it, it would be great to search the docs for "equipments" and fixing those
> 
> @williamjamir did you try this? There was only one incorrect word?

Yes, only one occurrence =)

docs/source/alfacase/03_alfacase_example.rst
williamjamir: Commit b3e9df5 improves the example section and fix the indentation errors =)

docs/source/alfacase/03_alfacase_example.rst
```diff
@@ -0,0 +1,396 @@
...
+.. note::
+
+    For the full reference check the :ref:`alfacase-reference-section`
+
+
+Project Options
+---------------
+
+We will start teh example by configuring the project options, as discussed in the previous section,
+the root level of an ``ALFacase`` file is the :class:`CaseDescription <alfasim_sdk.CaseDescription>` and it has the following options:
```
oliveira-mauricio, line 43: ALFacase with wrong casing, should be
```suggestion
the root level of an :program:`ALFAcase` file is the :class:`CaseDescription <alfasim_sdk.CaseDescription>` and it has the following options:
```
By the way, what does this :program: tag mean?

williamjamir: The short answer is font style.

The role `:program:` made more sense on the previous theme, which gave a special highlight for this name. But with this theme, it only applies a bold, which is kind doesn't look good in my opinion =/

PR #126, opened 2020-12-07
Add info about the default values on the definitions section

williamjamir:
ASIM-3584

Add the default value on the attribute in CaseDescription Tab

image

Add an "optional" label on the attribute (in schema Tab)
image

### DIFF
--- a/docs/source/alfacase_definitions/CaseOutputDescription.txt
+++ b/docs/source/alfacase_definitions/CaseOutputDescription.txt
@@ -5,22 +5,22 @@
     .. parsed-literal::
 
         class CaseOutputDescription
-            trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\]
-            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\
-            profiles: \ :class:`List <typing.List>`\[\ :class:`ProfileOutputDescription <ProfileOutputDescription>`\]
-            profile_frequency: \ :class:`Scalar <barril.units.Scalar>`\
+            trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\] = []
+            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(0.1, 's', 'time')
+            profiles: \ :class:`List <typing.List>`\[\ :class:`ProfileOutputDescription <ProfileOutputDescription>`\] = []
+            profile_frequency: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(0.1, 's', 'time')
 
 .. tab:: Schema
 
     .. parsed-literal::
 
             trends: 
-                - \ :class:`trend_output_description_schema <TrendOutputDescription>`\
+                - \ :class:`trend_output_description_schema <TrendOutputDescription>`\ # optional
             trend_frequency: 
                 value: number
-                unit: string
+                unit: string # optional
             profiles: 
-                - \ :class:`profile_output_description_schema <ProfileOutputDescription>`\
+                - \ :class:`profile_output_description_schema <ProfileOutputDescription>`\ # optional
             profile_frequency: 
                 value: number
-                unit: string
+                unit: string # optional
--- a/docs/source/alfacase_definitions/CaseOutputDescription.txt
+++ b/docs/source/alfacase_definitions/CaseOutputDescription.txt
@@ -6,21 +6,21 @@
 
         class CaseOutputDescription
             trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\]
-            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\ |space|
+            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\
             profiles: \ :class:`List <typing.List>`\[\ :class:`ProfileOutputDescription <ProfileOutputDescription>`\]
-            profile_frequency: \ :class:`Scalar <barril.units.Scalar>`\ |space|
+            profile_frequency: \ :class:`Scalar <barril.units.Scalar>`\
 
 .. tab:: Schema
 
     .. parsed-literal::
 
             trends: 
-                - \ :class:`trend_output_description_schema <TrendOutputDescription>`\ |space|
+                - \ :class:`trend_output_description_schema <TrendOutputDescription>`\
             trend_frequency: 
                 value: number
                 unit: string
             profiles: 
-                - \ :class:`profile_output_description_schema <ProfileOutputDescription>`\ |space|
+                - \ :class:`profile_output_description_schema <ProfileOutputDescription>`\
             profile_frequency: 
                 value: number
                 unit: string
--- a/docs/source/alfacase_definitions/CasingSectionDescription.txt
+++ b/docs/source/alfacase_definitions/CasingSectionDescription.txt
@@ -6,16 +6,16 @@
 
         class CasingSectionDescription
             name: str
-            hanger_depth: \ :class:`Scalar <barril.units.Scalar>`\
-            settings_depth: \ :class:`Scalar <barril.units.Scalar>`\
-            hole_diameter: \ :class:`Scalar <barril.units.Scalar>`\
-            outer_diameter: \ :class:`Scalar <barril.units.Scalar>`\
-            inner_diameter: \ :class:`Scalar <barril.units.Scalar>`\
-            inner_roughness: \ :class:`Scalar <barril.units.Scalar>`\
-            material: \ :class:`Optional <typing.Optional>`\[str]
-            top_of_filler: \ :class:`Scalar <barril.units.Scalar>`\
-            filler_material: \ :class:`Optional <typing.Optional>`\[str]
-            material_above_filler: \ :class:`Optional <typing.Optional>`\[str]
+            hanger_depth: \ :class:`Scalar <barril.units.Scalar>`
+            settings_depth: \ :class:`Scalar <barril.units.Scalar>`
+            hole_diameter: \ :class:`Scalar <barril.units.Scalar>`
+            outer_diameter: \ :class:`Scalar <barril.units.Scalar>`
+            inner_diameter: \ :class:`Scalar <barril.units.Scalar>`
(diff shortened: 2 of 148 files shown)


src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -83,7 +111,21 @@ def _get_declaration(
             for predicate_function, handle_function in list_of_predicate_and_handles:
                 if predicate_function(value.type):
                     attribute_value = handle_function(value.type)
-                    lines.append(f"{BASE_INDENT}{attribute_name}: {attribute_value}")
+                    default_value = _get_default_value_of_case(value)
+
+                    # For CaseDescription, if the attributes has a default value, the empty space is not necessary
+                    if default_value:
+                        attribute_value = attribute_value.replace("|space|", "")
```
oliveira-mauricio, line 118: What is the |space| for?

docs/source/alfacase_definitions/CaseOutputDescription.txt
```diff
@@ -5,22 +5,22 @@
     .. parsed-literal::
 
         class CaseOutputDescription
-            trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\]
-            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\ |space|
-            profiles: \ :class:`List <typing.List>`\[\ :class:`ProfileOutputDescription <ProfileOutputDescription>`\]
-            profile_frequency: \ :class:`Scalar <barril.units.Scalar>`\ |space|
+            trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\] = []
```
nicoddemus, line 8: This was auto-generated correct? Can you split the commit where you change the generator from the auto-generation, to make it easier for us to review it?

williamjamir: Sure

This commit changes the generator:  9dd2855
This commit updates the alfacase-definitions: 3f42885

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: code-block doesn't accept links or references, so I had to uses `literal-block` with " ` " characters
This brings some problems when the line ends with these special characters because sphinx doesn't insert a line break (I don't know the reason 🤷 )
My workaround solution was to add a space, but a normal space didn't work because pycharm was removing it 🤦 
So the solution was to uses a Unicode character that represents the space character.
The `| |` is just a way in sphinx re-use a text

Below is the docstring of `_get_class_with_reference` were wrote a docstring about this.
What your thoughts on this, this docstring is good enough?

```python
    :param bool add_space:
        inform if the name generated will need to have an unicode space at the end (generate by |space|),
        this is necessary when the class is the only reference of an attribute causing this line
        to be mixed with the next line over the documentation.

File: src/_alfasim_sdk/alfacase/generate_case_description_docstring.py [hunk 3]
```diff
@@ -60,21 +61,48 @@ def generate_definition(class_name: str) -> str:
...
 
 def _generate_declaration_for_schema(class_: Any) -> List[str]:
     """ Return all attributes for the given Class using ALFACase schema definition. """
     class_fields = attr.fields_dict(class_)
-    lines = _get_declaration(class_fields, LIST_OF_CASE_SCHEMAS)
-    return lines
+    return _get_declaration(class_fields, LIST_OF_CASE_SCHEMAS, is_schema=True)
+
+
+def _get_default_value_of_case(value: attr.Attribute) -> str:
```
nicoddemus, line 76: This name is a bit strange, how about `_get_case_attr_default_value`?

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -60,21 +61,48 @@ def generate_definition(class_name: str) -> str:
...
+    if isinstance(value.default, enum.Enum):
+        default_value_string += (
+            f"{value.default.__class__.__name__}.{value.default.name}"
+        )
+    elif isinstance(value.default, attr.Factory):
+        if value.default.factory is dict:
+            default_value_string += "{}"
+        elif value.default.factory is list:
+            default_value_string += "[]"
+        elif is_attrs(value.default):
```
nicoddemus, line 92: This `if`/`elif` chain doesn't have an `else` clause, which will produce a strange `" = "` string when we get a factory which doesn't match any of the ones being checked here. How about `else: default_value_string += "UNKNOWN FACTORY"` or something like this? While an ugly name, it will help us spot those in the docs and update accordingly.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
nicoddemus: Thanks for the explanation, but where do you get `'` from, from the default values? Can't we replace them by `"` instead during the default values generation? If that works, seems like a better solution to me as that would be a localized "hack" rather than something that spills through many parts of the code.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -83,7 +111,21 @@ def _get_declaration(
...
+                    default_value = _get_default_value_of_case(value)
+
+                    # For CaseDescription, if the attributes has a default value, the empty space is not necessary
+                    if default_value:
+                        attribute_value = attribute_value.replace("|space|", "")
+
+                    # For Schema,  if the attributes has a default value, the entry is optional
+                    if default_value and is_schema:
+                        attribute_value = attribute_value.replace("# optional", "")
+                        attribute_value += " # optional"
```
oliveira-mauricio, line 123: If your intent is just to add an extra space before `# optional` you could just:
>  attribute_value = attribute_value.replace("# optional", " # optional")

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
oliveira-mauricio: `_get_class_with_reference` is not even called directly in this function, so that docstring is far away from this code. Maybe just add a comment where you mention the empty spaces like `see _get_class_with_reference docs about the space`. The params doc itself is a bit confusing (the phrasing looks like the space will cause the lines to be mixed instead of preventing it), what about:

```
    :param bool add_space:
        inform if the name generated will need to have an unicode space at the end (generated by |space|),
        this is necessary when the class is the only reference of an attribute to avoid this line
        to be mixed with the next line over the documentation.
```

docs/source/alfacase_definitions/CasingSectionDescription.txt
```diff
@@ -12,10 +12,10 @@
             outer_diameter: \ :class:`Scalar <barril.units.Scalar>`\ |space|
             inner_diameter: \ :class:`Scalar <barril.units.Scalar>`\ |space|
             inner_roughness: \ :class:`Scalar <barril.units.Scalar>`\ |space|
```
prusse-martin, line 14: Did you try changing
```
           inner_roughness: \ :class:`Scalar <barril.units.Scalar>`\ |space|
```
to
```
           inner_roughness: \ :class:`Scalar <barril.units.Scalar>`
```
BTW, I assume `\` is the escape character.
Why are you escaping those spaces?
It that is really the escape character merging lines is what happens when you escape the newline character.
What will happen if you just remove all those escapes? If that is layout you are already using some custom css.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Nops, I'm adding `#optional` to the fields that have a default value.

Previously this `optional` label was being added only on fields that were `Optional[Something]`.
Perhaps I can change that since all options that are `Optional[Something]` have a default value.

docs/source/alfacase_definitions/CasingSectionDescription.txt
williamjamir: As we discussed on chat, commit 9dfefcd fix this issue.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: As we discussed on chat, commit 9dfefcd fix this issue.
(Removing the need of add_space or |space|)

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Sure

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Done

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Done on 22af421

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -83,72 +113,72 @@ def _get_declaration(
...
+
+                    # If attribute_value ends with backslash (because of the cross-reference), add extra space
+                    # for a default value.
+                    # Otherwise, the equal symbol will be sticky with the link.
+                    if default_value and attribute_value[-1] == "\\":
+                        default_value = f" {default_value}"
+
+                    line_content = f"{BASE_INDENT}{attribute_name}: {attribute_value}{default_value}"
+
+                    # Remove the last character if it's a backslash otherwise the line break will be avoided.
```
nicoddemus, line 131: This entire block of code is more complex/cryptic than it should be. 😁 

I suggest:

```python
# Remove the last character if it's a backslash otherwise the line break will be avoided.
if line_content.endswith("\\"):
    line_content = line_content[:-1]
lines.append(line_content) 
```

williamjamir: dammit, I thought that I was being clever 😅

nicoddemus (approved):

oliveira-mauricio (approved):
Shit it!

nicoddemus:
@williamjamir please make sure to squash/clean up the commits before merging.

PR #128, opened 2020-12-08
Add unit on doc

williamjamir:
Commit 273d671 adds the generator of the list of units
Commit 8a5ce20 updates tests and the txt with the content

Link for the demo: https://alfasim-sdk--128.org.readthedocs.build/en/128/

image

image

### DIFF
--- a/src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
+++ b/src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
@@ -33,6 +33,57 @@ INDENT = "    "
 """ Space needed to reach the root of the literal block. """
 BASE_INDENT = " " * 12
 
+CATEGORIES_USED_ON_DESCRIPTION = sorted(
+    [
+        "angle per time",
+        "density",
+        "dimensionless",
+        "dynamic viscosity",
+        "emissivity",
+        "flow coefficient",
+        "heat transfer coefficient",
+        "length",
+        "mass flow rate",
+        "mass fraction",
+        "mass per mol",
+        "molar thermodynamic energy",
+        "molar volume",
+        "mole per mole",
+        "power",
+        "pressure",
+        "productivity index",
+        "specific heat capacity",
+        "standard volume per standard volume",
+        "standard volume per standard volume",
+        "standard volume per time",
+        "temperature",
+        "thermal conductivity",
+        "time",
+        "velocity",
+        "volume flow rate",
+        "volume fraction",
+        "volumetric thermal expansion",
+    ]
+)
+
+
+def generate_list_of_units(category: str) -> str:
+    """ Return an admonition with toggle to show the units available for a given category. """
+    from barril.units import UnitDatabase
+
+    unit_database = UnitDatabase.GetSingleton()
+    units = unit_database.GetCategoryInfo(category).valid_units_set
+    info = unit_database.unit_to_unit_info
+    body = [f'    :"{unit}": {info[unit].name}' for unit in units]
+    lines = [
+        f".. admonition:: Available units for category '{category}'",
+        "    :class: dropdown",
+        "",
+        *sorted(body, key=str.casefold),
+        "",
+    ]
+    return "\n".join(lines)
+
 
 def generate_definition(class_name: str) -> str:
     """
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,8 @@
 exclude: |
     (?x)(
         ^docs/source/alfacase_definitions/|
-        ^tests/alfacase/test_generate_case_description_docstring/
+        ^tests/alfacase/test_generate_case_description_docstring/|
+        ^tests/alfacase/test_generate_list_of_units/
     )
 repos:
 -   repo: https://github.com/asottile/reorder_python_imports
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,3 +3,4 @@ sphinx-click==2.5.0
 pydata-sphinx-theme
 sphinx-inline-tabs
 sphinx-copybutton
+sphinx_togglebutton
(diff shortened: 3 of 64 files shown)


src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -33,6 +33,56 @@
...
+]
+
+
+def generate_list_of_units(category: str) -> str:
+    """ Return an admonition with toggle to show the units available for a given category. """
+    from barril.units import UnitDatabase
+
+    unit_database = UnitDatabase.GetSingleton()
+    title = unit_database.GetCategoryInfo(category).caption.title()
+    units = unit_database.GetCategoryInfo(category).valid_units_set
```
nicoddemus, line 74: Sort this please (in case `valid_units_set` is a `set`) to ensure a consistent order.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -33,6 +33,56 @@
 """ Space needed to reach the root of the literal block. """
 BASE_INDENT = " " * 12
 
+CATEGORIES_USED_ON_DESCRIPTION = [
```
nicoddemus, line 36: I believe this list should be sorted (probably easier to just call `.sort()` after the list definition).

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -33,6 +33,56 @@
...
+    """ Return an admonition with toggle to show the units available for a given category. """
+    from barril.units import UnitDatabase
+
+    unit_database = UnitDatabase.GetSingleton()
+    title = unit_database.GetCategoryInfo(category).caption.title()
+    units = unit_database.GetCategoryInfo(category).valid_units_set
+    info = unit_database.unit_to_unit_info
+    body = [f"    :{unit}: {info[unit].name}" for unit in units]
+    lines = [
+        f".. admonition:: Available units for category '{title}'",
```
nicoddemus, line 78: While I understand the reasoning, I don't think we should use `title` here: the user is supposed to write `"time"`, not `"Time"`, so using `'Time'` here is confusing.

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
```diff
@@ -33,6 +33,56 @@
...
+
+def generate_list_of_units(category: str) -> str:
+    """ Return an admonition with toggle to show the units available for a given category. """
+    from barril.units import UnitDatabase
+
+    unit_database = UnitDatabase.GetSingleton()
+    title = unit_database.GetCategoryInfo(category).caption.title()
+    units = unit_database.GetCategoryInfo(category).valid_units_set
+    info = unit_database.unit_to_unit_info
+    body = [f"    :{unit}: {info[unit].name}" for unit in units]
```
nicoddemus, line 76: Perhaps showing this as string/code would be better? 

```
    body = [f"    ``\"{unit}\"``: {info[unit].name}" for unit in units]
```

nicoddemus (commented):
Nice work, please see my comments!

prusse-martin (approved):

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: line 81 is already sorting the values =)
```
sorted(body, key=str.casefold)

Comment (nicoddemus) [reply src/_alfasim_sdk/alfacase/generate_case_description_docstring.py:75]: Ahh missed that, thanks!

File: src/_alfasim_sdk/alfacase/generate_case_description_docstring.py (hunk 3 above)
Comment (nicoddemus) [reply src/_alfasim_sdk/alfacase/generate_case_description_docstring.py]: I noticed you changed the code, but the docs are still showing the category as a title:

image

Did you forget to run the script that generates the docs?

File: src/_alfasim_sdk/alfacase/generate_case_description_docstring.py [hunk 5]
```diff
@@ -33,6 +33,57 @@
...
+    ]
+)
+
+
+def generate_list_of_units(category: str) -> str:
+    """ Return an admonition with toggle to show the units available for a given category. """
+    from barril.units import UnitDatabase
+
+    unit_database = UnitDatabase.GetSingleton()
+    units = unit_database.GetCategoryInfo(category).valid_units_set
```
oliveira-mauricio, line 75: Looking at the generated docs, I see that all units are being listed, while ALFAsim doesn't work with all of them (defined in `alfasim_gui/source/python/alfasim_gui/units.py` `RegisterUnits()`)

Example:
```length_units = ['m', 'ft', 'km', 'mm', 'in']```

Not sure what will happen if user uses one of the other units for length. If it works maybe there is no problem to list them all, could you check?

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Exactly, I forgot to run `inv cog` and the CI was not catching changes on the `alfacase_definitions` folder
Commit 221c7d4 adds this check and commit 6aa24a7 update the list ;)

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: hmm so this is weird, I thought that all configurations on barril were being made on alfasim_core, and I moved that file to alfasim-SDK.
I didn't know these configurations were spread across the application =/
So we need to move the  `alfasim_gui/units.py` to alfasim-SDK as well.

oliveira-mauricio:
Also noticed this in the preview (unrelated):

image

See, the last variable is messed up :/

src/_alfasim_sdk/alfacase/generate_case_description_docstring.py
williamjamir: Commit fd3668b move the content from alfasim_gui/units to alfasim_sdk

nicoddemus (approved):

oliveira-mauricio (approved):
Nice job.

PR #131, opened 2020-12-16
Fix units for volumetric flow rate at standard conditions

arthursoprana:
Change units (and category) of common testing input data

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,11 @@
 History
 =======
 
+UNRELEASED
+==================
+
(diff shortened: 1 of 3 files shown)


williamjamir:
I think that we don't need to update the changelog, but in the future, we need to be aware of these changes and bump the SDK version since it's a breaking change.

williamjamir (commented):
Update the changelog, since this is a breaking change 🥇

oliveira-mauricio (approved):

PR #132, opened 2021-01-12
Remove B_parameter

danilomendesdias:
This parameter is no longer required as we are dropping support for the respective method on PVT side

(RFDAP-2502)

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,7 @@ History
 * Add context support on ``alfasim_configure_fields``, ``alfasim_configure_layers`` and ``alfasim_configure_phases``.
 * Change category for ``volumetric_flow_rates_std` from ``volume flow rate`` to ``standard volume per time``.
 * Rename ``convert_alfacase_to_case`` to ``convert_alfacase_to_description``
+* Drop ``B_parameter`` as Lee-Chien method for surface tension is not supported anymore.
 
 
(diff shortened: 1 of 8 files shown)


nicoddemus (changes requested):
Please update the CHANGELOG as well.

danilomendesdias:
> Please update the CHANGELOG as well.

would be this a 0.9.0?

danilomendesdias:
> > Please update the CHANGELOG as well.
> 
> would be this a version 0.9.0?

Oh, the 0.8.0 is not published yet, so maybe I don't need to change that, right?

williamjamir (approved):
Awesome!
Thank you for taking the time to update this for us xD! 

I would normally comment that we need to keep backward compatibility, but this property is really particular; therefore we can drop it without weighing on our conscience 🙃

nicoddemus:
@williamjamir do we have some version _in the file format_ that we need to update?

williamjamir:
> @williamjamir do we have some version _in the file format_ that we need to update?
We have a file `version.py` on `alfasim_sdk.src._alfasim_sdk`, but this PR doesn't need to update that file because is already on `v0.8.0` (unreleased) and the last version at PyPI is `v0.7.0`

danilomendesdias:
ping @nicoddemus

nicoddemus (approved):

PR #133, opened 2021-01-15
Register gas standard volume per time category ASIM-2630

oliveira-mauricio:


williamjamir:
Perhaps we should add a new entry on `HISTORY.rst` informing this.

nicoddemus (approved):

williamjamir (approved):

PR #152, opened 2021-04-15
Fb asim 1780 add dt and per pipe trends v2

arthursoprana:


src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
```diff
@@ -192,7 +195,7 @@ def _get_class_with_reference(visible_name: str, ref: str) -> str:
     """
     Return the name of the class with a valid reference to be used by sphinx.
     """
-    return f"\\ :class:`{visible_name} <{ref}>`\\"
+    return f"\\ :class:`{visible_name} <{ref}>`"
```
williamjamir, line 198: I believe that this scape character at the end is necessary, otherwise, the docs are not formatted correctly (see the screenshot).
```suggestion
    return f"\\ :class:`{visible_name} <{ref}>`\\"
```
image

docs/source/alfacase_definitions/CaseOutputDescription.txt
```diff
@@ -7,28 +7,28 @@
         class CaseOutputDescription
             automatic_trend_frequency: bool = True
             automatic_trend_frequency: bool = True
-            trends: \ :class:`List <typing.List>`\[\ :class:`TrendOutputDescription <TrendOutputDescription>`\] = []
-            trend_frequency: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(0.1, 's', 'time')
+            trends: \ :class:`List <typing.List>`[\ :class:`TrendOutputDescription <TrendOutputDescription>`] = []
```
williamjamir, line 10: Nops the escape character is needed, see:
image

docs/source/alfacase_definitions/AnnulusDescription.txt
```diff
@@ -19,7 +19,7 @@
             has_annulus_flow: boolean
             has_annulus_flow: number
             pvt_model: string # optional
-            initial_conditions: \ :class:`initial_conditions_description_schema <InitialConditionsDescription>`\ # optional
-            gas_lift_valve_equipment: 
-                string: \ :class:`gas_lift_valve_equipment_description_schema <GasLiftValveEquipmentDescription>`\ # optional
+            initial_conditions: \ :class:`initial_conditions_description_schema <InitialConditionsDescription>` # optional
```
williamjamir, line 22: Please review these escape characters, since it's making some attributes to be missing on docs.
image

williamjamir (changes requested):
Please take a look at the escape character which is making the classes reference to be missing.

src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
```diff
@@ -169,7 +169,10 @@ def _get_declaration(
 
                     # For Schema: if the attributes have a default value the entry is optional.
                     if default_value and is_schema:
-                        attribute_value += " # optional"
+                        if attribute_value.startswith("\n"):
+                            attribute_value = "# optional" + attribute_value
+                        else:
+                            attribute_value += " # optional"
```
prusse-martin, line 175:
```suggestion
                        if attribute_value.startswith("\n"):
                            attribute_value = " # optional" + attribute_value
                        else:
                            attribute_value += "  # optional"
```
Just to avoid the `# optional` comment "touching" the attribute try to add a space here...

src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
prusse-martin: We should have a test checking the rendered stuff.
Or is that impractical?

williamjamir: IMHO, we already have a render test, just look at the number of files that were changed.
The issue here was a modification that affected several files and we didn't check the output, just ran `--force-regen`.

A visual render test is possible but in my opinion, would lead to the same outcome (executing `--force-regen` on a massive change), but now with more dependencies like starting a browser and so on.

What are your thoughts on this @oliveria, @bruno.

prusse-martin: Yes, we will force regen...
But the faulty result will be visible in the PR and not force someone to check the full generated documentation and manually compare it with the "master branch" one.
I am not saying we need to test all rendered files, just some, one for each "hack" we need top get the proper final rendering.

williamjamir (approved):

PR #153, opened 2021-04-16
Fb asim 4157 separator efficiency

edersongrein:


src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
```diff
@@ -954,6 +954,8 @@ def load_separator_node_properties_description(
         "initial_phase_volume_fractions": get_dict_with_scalar_loader(
             category="volume fraction"
         ),
+        "gas_separation_efficiency": get_scalar_loader(from_unit="%"),
```
ggrbill, line 957: I believe you have to make something similar in the `case_to_alfacase.py` file.
@williamjamir could you confirm?

ggrbill: Also, would be nice to add tests for the new entries.

ggrbill (changes requested):

src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
ggrbill: @williamjamir, I do not remember, do we need to add tests? Or the test suit makes it automatically for any new entry?

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1097,6 +1097,8 @@ class SeparatorNodePropertiesDescription:
             constants.FLUID_OIL: Scalar("volume fraction", 0.5, "-"),
         }
     )
+    gas_separation_efficiency = attrib_scalar(default=Scalar(1.0, "-"))
+    liquid_separation_efficiency = attrib_scalar(default=Scalar(1.0, "-"))
```
williamjamir, line 1101: I particularly think it's better to put the category explicitly and not rely on the default value (if I'm not mistaken is dimensionless)

src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
prusse-martin: No. `case_to_alfacase` is almost automatic.
I hope in the future to mate a bit less reliant o manual updates =)

williamjamir: > I believe you have to make something similar in the case_to_alfacase.py file.

Case to ALFAcase is more "smart", is just a loop over all attributes, and automatically gets the value.
You will need to update that only if a new "type" is added.

> Also, would be nice to add tests for the new entries.
Definitely, please update the `SeparatorNodePropertiesDescription` on `alfasim-sdk/tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py` and add these attributes there as well.

> @williamjamir, I do not remember, do we need to add tests? Or the test suit makes it automatically for any new entry?

There is a basic "round-trip" test that takes a filled case description and converts it to ALFAcase and reimport to check if the values match.
With this test, you can validate that the conversion (ALFAcase->case; case->ALFAcase) is working.

ggrbill: Thanks a lot, Guys!

@edersongrein if you don't understand some of the instructions, please let me know, I can help you on that.

williamjamir: We didn't add (yet) a smart way to automatically check if the filled case description is outdated (that would be nice).

I think that one possible solution is to add a "template" when a new PR is opened with TODOS to be check and one of the options is "Did you update the filled case description available at X ? ".

ggrbill: Sounds good to me. I think it would help.

williamjamir (changes requested):
Please update the `SeparatorNodePropertiesDescriptio` class on
`alfasim-sdk/tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py`

Also, you need to run the tests with `force-regen` to update the expected values (this is the reason that the CI is failing).

src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
edersongrein: > Thanks a lot, Guys!
> 
> @edersongrein if you don't understand some of the instructions, please let me know, I can help you on that.

Thanks. Probably I will

src/alfasim_sdk/_internal/alfacase/case_description.py
edersongrein: Done

ggrbill (approved):

williamjamir (approved):
Awesome! 🚀 

Just don't forget to update the HISTORY.rst with these new additions ;)

HISTORY.rst
```diff
@@ -5,7 +5,7 @@ History
 0.9.0 (unreleased)
 ==================
 
-*
+* Add gas and liquid separation efficiency to ``Separator`` model
```
prusse-martin, line 8:
```suggestion
* Add gas and liquid separation efficiency to ``Separator`` model.
```

prusse-martin (approved):

PR #156, opened 2021-04-20
Add Hook for Emulsion Model

ggrbill:


docs/source/plugins/06_solver_hooks.rst
```diff
@@ -237,6 +237,16 @@ each Liquid-Liquid system plugin hooks. In the sequence, they are presented
 
 .. autofunction:: alfasim_sdk._internal.hook_specs.calculate_liq_liq_shear_force_per_volume
 
+Emulsion Model
+-------------------------------
+
+When a simulation case uses a three-phase flow, the |alfasim|'s simulator allows to select an emulsion model that will
```
edersongrein, line 243: I'm not sure if this is what you meant, but what about this rephrase:

"When a simulation case uses a three-phase flow model, |alfasim| allows the selection of an emulsion model to calculate emulsion viscosity. The model may be further customized through the implementation of a plugin for relative emulsion viscosity."

arthursoprana (approved):

docs/source/plugins/06_solver_hooks.rst
ggrbill: I think your suggestion is very good, but I want to give the information on how this hook is used, which is `update the layer viscosity related to the emulsion`.

edersongrein: Ok, no problem

edersongrein (approved):

docs/source/plugins/06_solver_hooks.rst
ggrbill: I updated it as we discussed in the RC.

prusse-martin (commented):
=/
I forgot to hit the final "Submit review"  button.

src/alfasim_sdk/_internal/hook_specs.py
```diff
@@ -787,7 +782,7 @@ def finalize_state_variables_calculator(ctx: "void*") -> "int":
         :linenos:
         :emphasize-lines: 1
 
-        HOOK_FINALIZE_STATE_VARIABLES_CALCULATOR(void* ctx)
+        HOOK_FINALIZE_STATE_VARIABLES_CALCULATOR(ctx)
```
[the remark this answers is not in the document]
ggrbill: It is OK. I mean, when the `alfasim-sdk template` generates a plugin, it adds macros in the `hook_spec.h` in which it is supposed to use without the type specification, since it is already added in the macro.

I removed it because I realize that when plugin developers start to implement one of those hooks, the first thing they do is copy/paste from the documentation. Since it is wrong to use the macro with type specification it can cause a compile error.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1996,6 +1996,9 @@ class PhysicsDescription:
     )
     keep_former_results: bool = attr.ib(default=False, validator=instance_of(bool))
     emulsion_model = attrib_enum(default=constants.EmulsionModelType.NoModel)
+    emulsion_model_plugin_id: str = attr.ib(
+        default="", validator=optional(instance_of(str))
```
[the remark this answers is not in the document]
ggrbill: Hummm, my bad. This property should not be optional.
I will open a PR correcting those things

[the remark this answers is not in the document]
ggrbill: Fixed in PR #163

PR #160, opened 2021-04-26
Add new CLI command called 'update'

ggrbill:
This command will update any automatically generated files
by alfasim-sdk.

It is important when a plugin must be updated to work properly
with the new alfasim-sdk version.

For now, only `hook_specs.h` is being updated, bu more files can be added in the future.


src/alfasim_sdk/_internal/cli.py
```diff
@@ -184,6 +184,37 @@ def package_only(ctx, plugin_dir, package_name, dst):
...
+    ALFAsim-SDK is released. Otherwise compilation problems can occur.
+
+    The files updated are:
+    - ´<plugin_folder>/src/hook_specs.h´
+    """
+    hook_specs_file_path = _get_hook_specs_file_path()
+    hm = HookManGenerator(hook_spec_file_path=hook_specs_file_path)
+
+    plugin_folder = Path(plugin_dir)
+    hook_specs_h_folder = plugin_folder / "src" / "hook_specs.h"
```
ggrbill, line 206: - [ ] `hook_specs_h_path`

src/alfasim_sdk/_internal/cli.py
```diff
@@ -184,6 +184,37 @@ def package_only(ctx, plugin_dir, package_name, dst):
...
+
+    if not hook_specs_h_folder.exists() or not hook_specs_h_folder.is_file():
```
ggrbill, line 208:
```suggestion
    if not hook_specs_h_path.exists() or not hook_specs_h_path.is_file():
```

tests/test_cli.py
```diff
@@ -94,3 +94,21 @@ def test_compile_command(tmp_path):
...
+    )
+
+    plugin_src_folder = tmp_path / "plugin_a" / "src"
+    plugin_src_folder.mkdir(parents=True)
+    (plugin_src_folder / "hook_specs.h").write_text(data="", encoding="utf-8")
+    result = runner.invoke(
+        console_main, ["update", "--plugin-dir", plugin_src_folder.parent]
+    )
+    assert result.exception is None
+    assert result.output == ""
```
prusse-martin, line 114: You could check that `(plugin_src_folder / "hook_specs.h")` is no longer empty.

prusse-martin (changes requested):

tests/test_cli.py
ggrbill: It is already tested in another place when the function that generates the file is implemented.
Here, only the CLI commands are tested. See the other CLI commands Tests.

ggrbill: Although, I will check as simple as possible, like `size > 0`

- [ ] check if file is not empty

prusse-martin (approved):

src/alfasim_sdk/_internal/cli.py
```diff
@@ -184,6 +184,37 @@ def package_only(ctx, plugin_dir, package_name, dst):
     )
 
 
+@console_main.command()
+@click.pass_context
+@plugin_dir_option
+def update(ctx, plugin_dir):
```
nicoddemus, line 190: We already have a `cog` task which does something similar. Should we merge them together perhaps?

ggrbill: I am sorry. I didn't follow you. 

The `cog` task is not related to this new command that I am adding to the CLI.

src/alfasim_sdk/_internal/cli.py
```diff
@@ -184,6 +184,37 @@ def package_only(ctx, plugin_dir, package_name, dst):
...
+    """
+    Updates plugin files automatically generated by ALFAsim-SDK.
+    The plugin folder is informed with plugin-dir option.
+    When not provided plugin-dir will be the current folder location.
+
+    It is important to update the plugin automatically generated files when a new version of
```
arthursoprana, line 196:
```suggestion
    It is important to update the automatically generated files from each plugin when a new version of
```

arthursoprana (approved):

src/alfasim_sdk/_internal/cli.py
ggrbill: Done

src/alfasim_sdk/_internal/cli.py
nicoddemus: I meant both tasks get the source code "ready" to use the library. We should merge the two, otherwise the user needs to know which task to execute to update the source.

Could have asked for clarification before hitting merge.

prusse-martin: `inv cog` generates code for the sdk, is used in development of sdk it self to generate the code that will be used by client code.
`alfasim-sdk update` updates the client code of plugin authors using sdk.

nicoddemus: Yes my bad.

PR #161, opened 2021-04-28
Add tracers to initialize_state_variables_calculator signature

arthursoprana:
This information may be used by plugins, it's a prototype implementation.

### NOTE: This is a work in progress prototype. Only to initialize discussions.


nicoddemus:
> NOTE: This is a work in progress prototype. Only to initialize discussions.

Just a tip: you can open a "draft" PR to convey that (I think it is even possible to convert this PR to draft after it being opened, but I'm not sure).

PR #164, opened 2021-04-30
Do not render bool attributes twice in description files

oliveira-mauricio:
- Break loop when first match is found to avoid duplicated attributes;
- Change is_int function to not consider booleans as ints to avoid relying on execution order to match bool/int;


nicoddemus (approved):

williamjamir (approved):

src/alfasim_sdk/_internal/alfacase/generate_schema.py
```diff
@@ -156,7 +156,8 @@ def array_to_alfacase_schema(type_, block_indentation):
 
 
 def is_int(type_):
-    return inspect.isclass(type_) and issubclass(type_, int)
+    # bool is a subclass of int, but we are not intersted in matching that here.
+    return inspect.isclass(type_) and issubclass(type_, int) and type_ is not bool
```
prusse-martin, line 160: Why not

```suggestion
    return inspect.isclass(type_) and issubclass(type_, int) and not is_boolean(type_)
```

src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
```diff
@@ -188,6 +188,7 @@ def _get_declaration(
                         line_content = line_content[:-1]
 
                     lines.append(line_content)
+                    break  # Match found, break inner loop
```
prusse-martin, line 191: 👍🏻

prusse-martin (approved):

PR #170, opened 2021-05-11
Account for changes in `attr` package

prusse-martin:
This should not cause problems with older versions of the `attr`
package.

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
+++ b/src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
@@ -143,7 +143,7 @@ def _get_case_attr_default_value(value: attr.Attribute) -> str:
             default_value_string += "{}"
         elif value.default.factory is list:
             default_value_string += "[]"
-        elif is_attrs(value.default):
+        elif is_attrs(value.default.factory):
             default_value_string += f"{value.default.factory.__name__}()"
         else:
             default_value_string += "UNKNOWN FACTORY"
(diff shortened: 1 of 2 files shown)


src/alfasim_sdk/_internal/alfacase/generate_case_description_docstring.py
```diff
@@ -143,7 +143,7 @@ def _get_case_attr_default_value(value: attr.Attribute) -> str:
             default_value_string += "{}"
         elif value.default.factory is list:
             default_value_string += "[]"
-        elif is_attrs(value.default):
+        elif is_attrs(value.default.factory):
```
prusse-martin, line 146: Formerly we where checking the `Factory` class not the actual factory used.
With older version of `attr` the `Factory` class will pass our `is_attrs` check, this is not true with recent versions.

williamjamir: Hmmm, weird looking at the `case_description.py` I could not find any references for Factories besides dicts or lists.
Do you know which field this statement is covering?

prusse-martin: [`WellDescription.profile = attrib_instance(ProfileDescription)`](https://github.com/ESSS/alfasim-sdk/blob/6f87fd21b539c4ac0dc99f8271f0a996ab0f644d/src/alfasim_sdk/_internal/alfacase/case_description.py#L1497)
The [`attrib_instance`](https://github.com/ESSS/alfasim-sdk/blob/6f87fd21b539c4ac0dc99f8271f0a996ab0f644d/src/alfasim_sdk/_internal/alfacase/case_description_attributes.py#L310) set the default to a Factory.

williamjamir (approved):

nicoddemus (approved):

PR #171, opened 2021-05-20
Update HOOK_INITIALIZE_STATE_VARIABLES_CALCULATOR

arthursoprana:


src/alfasim_sdk/_internal/hook_specs.py
```diff
@@ -526,12 +528,15 @@ def initialize_state_variables_calculator(
     :param P: Pressure values array
     :param T: Temperature values array
     :param T_mix: Mixture temperature values array
+    :param phi: Tracer mass fraction values array
```
ggrbill, line 531: Since this variable is the only one with different dimensions from the others, please add in the example of using this hook the way to get a mass fraction from a specific tracer.

ggrbill (changes requested):

src/alfasim_sdk/_internal/hook_specs.py
arthursoprana: I'm not sure if adding an example right now is a good idea. Mainly because I don't have a concrete working example that uses `phi`. I prefer to first create a working plugin and after that adding the example. My motivation is that things could change and we may forget to update the docs here.

PR #172, opened 2021-05-24
Review doc links

capmayer:


docs/source/plugins/02_alfasim_sdk_by_example.rst
```diff
@@ -218,7 +218,7 @@ needed to load and unload the |sdk| API, in which will allows the plugin to use
 Then, since the plugin wants to calculate its own secondary variable, the
 :func:`HOOK_UPDATE_PLUGINS_SECONDARY_VARIABLES <alfasim_sdk._internal.hook_specs.update_plugins_secondary_variables>` must be implemented.
 As can be seen in the example below, to retrieve the velocity of the continuous liquid field
-it is necessary to use the :func:`get_simulation_array` API function.
+it is necessary to use the ``get_simulation_array`` API function.
```
ggrbill, line 221: You are removing the link. You should only add `:cpp:` in front of it, like `:cpp:func:`.
In that way the link will be fixed.

src/alfasim_sdk/_internal/hook_specs_gui.py
```diff
@@ -40,7 +40,7 @@ def alfasim_get_data_model_type():
     .. |m_2| image:: /_static/images/marker_2.png
         :scale: 80%
 
-    |m_1| Location to where the models :func:`~alfasim_sdk.models.container_model` or :func:`~alfasim_sdk.models.data_model` will be placed. |br|
+    |m_1| Location to where the models :func:`alfasim_sdk.container_model` or :func:`alfasim_sdk.data_model` will be placed. |br|
```
ggrbill, line 43: Please, keep the `~` it will add a link with just the name of the structure instead of the entire path.

src/alfasim_sdk/_internal/hook_specs_gui.py
```diff
@@ -67,8 +67,8 @@ def alfasim_get_data_model_type():
 
     Example 2: This second example shows how to create a new container model.
 
-    Notice that when using the :func:`~alfasim_sdk.models.container_model` you only need to inform the container class
-    to the :func:`~alfasim_sdk.hook_specs_gui.alfasim_get_data_model_type` hook
+    Notice that when using the :func:`alfasim_sdk.container_model` you only need to inform the container class
+    to the :func:`alfasim_get_data_model_type` hook
```
ggrbill, line 71: Please keep `~`, like my others comments

src/alfasim_sdk/_internal/hook_specs_gui.py
```diff
@@ -173,10 +173,10 @@ def alfasim_get_status(
 
     The status monitor accepts two types of message:
 
-     - :func:`~alfasim_sdk.status.ErrorMessage`:
+    :class:`alfasim_sdk.ErrorMessage`:
```
ggrbill, line 176: If possible (I am not sure if the behavior is the same in `:class:`), add the `~` in front of the `alfasim_sdk.ErrorMessage`.

src/alfasim_sdk/_internal/status.py
```diff
@@ -11,9 +11,12 @@ class ErrorMessage:
     the simulation until the issue is fixed.
 
     :param model_name: Name of the model that issues the error.
+    :type model_name: str
+
     :param message: Message that will be displayed over the status monitor.
+    :type message: str
 
-    Checkout the :func:`~alfasim_sdk.hook_specs_gui.alfasim_get_status` for some examples of :func:`~alfasim_sdk.status.ErrorMessage` in action.
+    Checkout the :func:`alfasim_sdk._internal.hook_specs_gui.alfasim_get_status` for some examples of :class:`ErrorMessage` in action.
```
ggrbill, line 19: I think here is possible to use the same `~` at list for the `:func:` type.

src/alfasim_sdk/_internal/status.py
```diff
@@ -27,9 +30,12 @@ class WarningMessage:
     issue that needs to be fixed but doesn't block the simulation.
 
     :param model_name: Name of the model that issues the warning.
+    :type model_name: str
+
     :param message: Message that will be displayed over the status monitor.
+    :type message: str
 
-    Checkout the :func:`~alfasim_sdk.hook_specs_gui.alfasim_get_status` for some examples of :func:`~alfasim_sdk.status.WarningMessage` in action.
+    Checkout the :func:`alfasim_sdk._internal.hook_specs_gui.alfasim_get_status` for some examples of :class:`WarningMessage` in action.
```
ggrbill, line 38: Also here, use `~` at least for `:func:` type

ggrbill (changes requested):
Please, review all links that `~` was removed. The difference between using it and not is in the image below. The left one is your change, removing `~` and the right one is the original, using `~` in the beginning of the path.

image

docs/source/plugins/02_alfasim_sdk_by_example.rst
capmayer: Nice, added!

docs/source/conf.py
```diff
@@ -73,11 +73,11 @@
 .. |sdk| replace:: :program:`ALFAsim-SDK`
 
 .. |template-command| replace:: :ref:`Template Command <alfasim_sdk_cli_template_section>`
-.. |gui_hook| replace:: :py:func:`alfasim_get_data_model_type`
+.. |gui_hook| replace:: :py:func:`alfasim_sdk._internal.hook_specs_gui.alfasim_get_data_model_type`
 .. |container| replace:: :py:func:`alfasim_sdk.container_model`
 .. |model| replace:: :py:func:`alfasim_sdk.data_model`
 .. |s_variable| replace:: :py:func:`alfasim_sdk.SecondaryVariable`
-.. |s_variable_hook| replace:: :py:func:`alfasim_get_additional_variables`
+.. |s_variable_hook| replace:: :py:func:`alfasim_sdk._internal.hook_specs_gui.alfasim_get_additional_variables`
```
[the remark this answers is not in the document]
capmayer: Done!

ggrbill (approved):
Please, double check section `Plugin By Example` in the documentation. There are some links that have the entire path. 

See it is possible to fix them (I think there are 3 of them).

Good work!

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -131,26 +131,26 @@ class CaseOutputDescription:
 @attr.s(kw_only=True)
 class _MassSourceCommon:
     """
-    :ivar total_mass_flow_rate:
+    :param total_mass_flow_rate:
```
williamjamir, line 134: Was this change intentional to change the text in the documentation (from `Variable` to `Parameters`)? 

Because `:param:` is a well-known field for functions parameter, and perhaps we should stick with an explicit version `:parameter:`, so it will be clear that is this field is not related to a function parameter.

what are your thoughts on this @oliveira-mauricio @prusse-martin

docs/README.rst
```diff
@@ -0,0 +1,33 @@
+=========================
+alfasim-sdk documentation
+=========================
+
+**Some advices when creating the documentation.**
```
williamjamir, line 5:
```suggestion
**Some advice when creating the documentation.**
```

docs/README.rst
```diff
@@ -0,0 +1,33 @@
+=========================
```
williamjamir, line 1: Awesome! 
Really well-toughed file, a guide like this with our decisions is more than needed!

williamjamir (changes requested):
Awesome PR, thank you for taking the time to review the docs as well while fixing the links.

Let's discuss more the topics that I brought in my comments.

docs/Makefile
```diff
@@ -2,7 +2,7 @@
 #
 
 # You can set these variables from the command line.
-SPHINXOPTS    =
+SPHINXOPTS    = -W
```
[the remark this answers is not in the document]
capmayer: Hey, my bad, in windows, this file wasn't changing  anything. I will add `-W --keep-going` to this file and  `make.bat`

src/alfasim_sdk/_internal/types.py
```diff
@@ -210,9 +210,10 @@ class String(BaseField):
     """
     The String field represents an input that allows the user to enter and edit a single line of plain text.
 
-    The String field have all options available from :func:`~alfasim_sdk.types.BaseField`, plus the following ones
+    The String field have all options available from ```BaseField```, plus the following ones
```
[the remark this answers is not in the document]
capmayer: Yeah, it can't find the reference for BaseField. `WARNING: py:class reference target not found: BaseField`

I was searching about it, and probably we need to change the `99_2_user_interface_reference` to use `.. automodule::`

docs/source/alfacase/alfacase.example.rst
```diff
@@ -6,7 +6,7 @@ Example
 In this section, we will walk trough the creation of a simple network, with two nodes and one pipe, to illustrate how to
 crete a project and how to interpret the definitions available on :ref:`alfacase-reference-section`.
 
-The root level of our :program:`ALFAcase` file is always a :class:`CaseDescription <alfasim_sdk.CaseDescription>`
+The root level of our :program:`ALFAcase` file is always a :py:class:`CaseDescription <alfasim_sdk.CaseDescription>`
```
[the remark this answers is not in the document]
capmayer: No, I tried to add it to files that have `:cpp:` too, more for a visual identification.

docs/source/conf.py
```diff
@@ -131,3 +131,21 @@
...
+
+# -- Nitpicky config ---------------------------------------------------------
+
+nitpicky = True
+nitpick_ignore = [
+    ("py:class", "type"),
+    ("py:class", "barril.curve.curve.Curve"),  # Barril doesn't have Curve documented.
+    ("py:class", "Path"),
+    ("py:exc", "FrozenInstanceError"),
+    ("cpp:identifier", "ALFAsimSDK_API"),  # Must be checked.
```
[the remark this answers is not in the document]
capmayer: Some classes and functions used in docs can't be referred, so we add it here, like `Curve` from `barril` that don't have a public documentation.

I couldn't figure out the `("cpp:identifier", "ALFAsimSDK_API")`, any suggestion?

src/alfasim_sdk/_internal/alfacase/case_description.py
nicoddemus: Not sure I follow... the original intent was to document the instance variables right?

I prefer `:param:` over `:parameter:` because the former is shorter and there's really no confusion when using it.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -713,7 +713,9 @@ class TableIPRDescription(CommonIPR):
 @attr.s(frozen=True, slots=True)
 class IPRModelsDescription:
     """
-    :cvar Dict[str, Union[str, IPRDescription]] tables:
+    :type tables Dict[str, Union[str, IPRDescription]]
```
nicoddemus, line 716: I don't think we need this `:type` directive at all.

Where is `tables` declared? I only see `linear_models` and `table_models` below.

src/alfasim_sdk/_internal/context.py
```diff
@@ -65,14 +67,18 @@ class PipelineInfo:
     The PipelineInfo provides information about the geometry of a pipeline.
 
     PipelineSegmentInfo provides the following attributes:
-        name: Name associated with this ``Pipeline`` on ALFAsim
+    :param name: Name associated with this ``Pipeline`` on ALFAsim
+    :type name: str
 
-        edge_name: Name of the edge that this ``Pipeline`` is associated with.
+    :param edge_name: Name of the edge that this ``Pipeline`` is associated with.
+    :type edge_name: str
```
nicoddemus, line 74: All the `:param` and `:type` fixing done here (well done btw) was done automatically?

Just to comment that in the future I think it is better to use proper type annotations rather than `:type` directives.

src/alfasim_sdk/_internal/types.py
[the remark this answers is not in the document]
nicoddemus: For me it is declared in `alfasim_sdk._internal.types`, so I suppose it is just a matter of using

```
:func:`~alfasim_sdk._internal.types.BaseField`
```

?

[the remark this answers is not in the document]
capmayer: I tried it, but seems that sphinx can't refer to the base class this way.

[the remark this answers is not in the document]
capmayer: Hey, find out that the problem was that in `99_2_user_interface_reference.rst` BaseField was using the full path in `.. autoclass:: alfasim_sdk._internal.types.BaseField()`and the other classes are using the short version `.. autoclass:: alfasim_sdk.String()` (they are imported in __init__). Importing `BaseField`and using the short path solves the problem, now we can reference the base class.

src/alfasim_sdk/_internal/alfacase/case_description.py
capmayer: Ops, typo... 

We may need to choose a pattern for our docstring. By default `sphinx` uses `:type` for typing.
With the google pattern we could use the type annotations. 
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

capmayer:
```
def function_with_pep484_type_annotations(param1: int, param2: str) -> bool:
    """Example function with PEP 484 type annotations.

    Args:
        param1: The first parameter.
        param2: The second parameter.

    Returns:
        The return value. True for success, False otherwise.

    """

File: src/alfasim_sdk/_internal/alfacase/case_description.py (hunk 8 above)
Comment (oliveira-mauricio) [reply src/alfasim_sdk/_internal/alfacase/case_description.py:134]: I don't get the reason for changing this, what is the goal? Also, where is this in the online docs? I couldn't find it.

File: src/alfasim_sdk/_internal/context.py [hunk 17]
```diff
@@ -191,8 +196,8 @@ def alfasim_get_data_model_type():
             >>> ctx.get_model('MyModel').name
             'ALFAsim'
 
-        At runtime, you can also verify the names of the models defined by a given plugin. For this, you need to
```
oliveira-mauricio, line 194: Good catch.

src/alfasim_sdk/_internal/alfacase/case_description.py
nicoddemus: Sphinx standard also reads type annotations... that's what we use in pytest. 😁 

In summary, both Google and Sphinx notations read type annotations, so we can drop them from the docs. 👍

capmayer: Ohh cool! I take a look in pytest `conf.py` and there is used `autodoc_typehints = "description"`. I used it here and don't have the same effect. I'll review all types and try to remove then from docstring.

But looking at pytest doc, some places it don't have annotations in the docs:
https://docs.pytest.org/en/6.2.x/reference.html?highlight=mark#node

nicoddemus: >  I used it here and don't have the same effect. I'll review all types and try to remove then from docstring.

Hmm what do you mean? The first phrase conveys that it didn't work, but the second phrase to me indicates it did work and you will review the other types... can you clarify please?

> But looking at pytest doc, some places it don't have annotations in the docs

Probably they are not type-annotated yet, but don't worry about it.

capmayer: Yeah, it didn't work, but I will review all annotations and remove then from docs to see how it will behave.

nicoddemus (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
capmayer: Talking to @williamjamir I get more context about the documentation. 

It was my bad, as I saw `:param` being used in another class (i.e `PvtModelsDescription`) so I thought it would be great to have a pattern. Also using `:param` renders a `Parameters` section in the docs over a `Variables` (William mention in the comment above), witch I think is more intuitive for users that aren't exactly programmers.

Another point it that we have a lot of blank space using `:param` or `:ivar`: 
https://alfasim-sdk.readthedocs.io/en/latest/alfacase/api_reference.html?highlight=%20PvtModelsDescription#alfasim_sdk.PvtModelsDescription

I think we have a lot to debate about the docs, but for this PR I will stick to the plan of reviewing the links and removing types from docs.

nicoddemus: > I think we have a lot to debate about the docs, but for this PR I will stick to the plan of reviewing the links and removing types from docs.

I agree. 👍

oliveira-mauricio: That means you will change those back to ivar?

capmayer: Yeah, I changed to `:ivar` but now `sphinx` can't find the reference for those: `WARNING: py:obj reference target not found: correlations`

capmayer: With `:param` it doesn't show any warning

src/alfasim_sdk/_internal/alfacase/case_description.py
oliveira-mauricio: It's more a matter of concept than rendering... those attributes are instance vars, not parameters. If the rendering is not great/has warnings that's another problem to solve.

src/alfasim_sdk/_internal/alfacase/case_description.py
oliveira-mauricio: Just to register: this warning on :ivar: was fixed in a more recent version of sphinx, we will try to update it to solve this issue.

williamjamir (approved):

PR #174, opened 2021-06-02
Update relative emulsion viscosiry hook

ggrbill:
Since it is supposed to implement a correlation, now this hook receives double instead of an array of doubles.

### DIFF
--- a/docs/source/plugins/06_solver_hooks.rst
+++ b/docs/source/plugins/06_solver_hooks.rst
@@ -243,8 +243,8 @@ each Liquid-Liquid system plugin hooks. In the sequence, they are presented
     are implemented, the |alfasim| will show an error of configuration.
 
 .. note ::
-    Note that the all of the hooks related to the liquid-liquid system of Mechanistic Model can use same emulsion
-    viscosity model used internally by |alfasim|. It is made by helper function :cpp:func:`get_relative_emulsion_viscosity`.
+    Note that the all of the hooks related to the liquid-liquid system of Mechanistic Model can use the same emulsion
+    viscosity model used internally by |alfasim|. It can be obtained by using the helper function :cpp:func:`get_relative_emulsion_viscosity`.
     Since the liquid-liquid system is an emulsion, this helper function is important to keep the consistency between
     |alfasim| and its plugins that implement the following hooks.
 
@@ -268,8 +268,9 @@ the implementation of a relative emulsion viscosity correlation.
 .. autofunction:: alfasim_sdk._internal.hook_specs.calculate_relative_emulsion_viscosity
 
 .. warning::
-    The emulsion viscosity hook must be implemented as a simple correlation, in which all data is available as
-    parameters (Dispersed and continuous viscosities, for example). For this hook only the API functions
+    The relative emulsion viscosity hook has a limited API. Therefore the user-implemented correlations will have
+    access only to dispersed field viscosity, continuous field viscosity and volume fraction of the dispersed phase.
+    These variables are passed as arguments of the hook signature. For this hook only the API functions
     :cpp:func:`get_field_id`, :cpp:func:`get_phase_id` and :cpp:func:`get_layer_id` are available, any other API
     function will return an error code different from ``OK`` (:cpp:enum:`error_code` values) if called from this hook.
 
--- a/docs/source/plugins/06_solver_hooks.rst
+++ b/docs/source/plugins/06_solver_hooks.rst
@@ -240,7 +240,14 @@ each Liquid-Liquid system plugin hooks. In the sequence, they are presented
 
 .. warning::
     The hooks described in this section must be implemented all of them or none of them. If just part of four hooks
-    are implemented, the |Alfasim| will show an error of configuration.
+    are implemented, the |alfasim| will show an error of configuration.
+
+.. note ::
+    Note that the all of the hooks related to the liquid-liquid system of Mechanistic Model can use same emulsion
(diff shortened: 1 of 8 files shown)


docs/source/plugins/06_solver_hooks.rst
```diff
@@ -260,6 +267,12 @@ the implementation of a relative emulsion viscosity correlation.
 
 .. autofunction:: alfasim_sdk._internal.hook_specs.calculate_relative_emulsion_viscosity
 
+.. warning::
+    The emulsion viscosity hook must be implemented as a simple correlation, in which all data is available as
+    parameters (Dispersed and continuous viscosities, for example). For this hook only the API functions
```
edersongrein, line 272: Dispersed->dispersed

edersongrein (approved):

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -240,7 +240,14 @@ each Liquid-Liquid system plugin hooks. In the sequence, they are presented
 
 .. warning::
     The hooks described in this section must be implemented all of them or none of them. If just part of four hooks
-    are implemented, the |Alfasim| will show an error of configuration.
+    are implemented, the |alfasim| will show an error of configuration.
+
+.. note ::
+    Note that the all of the hooks related to the liquid-liquid system of Mechanistic Model can use same emulsion
```
arthursoprana, line 246:
```suggestion
    Note that the all of the hooks related to the liquid-liquid system of Mechanistic Model can use the same emulsion
```

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -240,7 +240,14 @@ each Liquid-Liquid system plugin hooks. In the sequence, they are presented
...
+    viscosity model used internally by |alfasim|. It is made by helper function :cpp:func:`get_relative_emulsion_viscosity`.
```
arthursoprana, line 247: "It is made by helper function :cpp:func:`get_relative_emulsion_viscosity`."

I didn't fully understood this sentence. Could you rephrase it? Is it supposed to be something like "The emulsion viscosity model can be obtained by using the helper function :cpp:func:`get_relative_emulsion_viscosity`."?

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -260,6 +267,12 @@ the implementation of a relative emulsion viscosity correlation.
 
 .. autofunction:: alfasim_sdk._internal.hook_specs.calculate_relative_emulsion_viscosity
 
+.. warning::
+    The emulsion viscosity hook must be implemented as a simple correlation, in which all data is available as
```
arthursoprana, line 271: I think we could avoid mentioning "must be implemented as a simple correlation", since they are not necessarily simple, just have a limited access to the API ;)

```suggestion
    The relative emulsion viscosity hook has a limited API. Therefore the user-implemented correlations will have access only to dispersed field viscosity, continuous field viscosity and volume fraction of the dispersed phase. These variables are passed as arguments of the hook signature.
```

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -778,4 +778,38 @@ DLL_EXPORT int get_liq_liq_shear_force_per_volume_input_variable(
...
+    During the implementation of any HOOK related to the Liquid-Liquid Mechanistic model, this
+    function provides the relative emulsion viscosity from Emulsion Model (selected through GUI).
+
+    It allows the plugin HOOKs to use the same emulsion model used internally by ALFAsim.
+
+    The definition of relative viscosity is given by the ratio between the apparent viscosity
+    (dispersed field + continuous field) and the continuous field.
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] out Relative Emulsion Viscosity [-].
```
arthursoprana, line 793: Is this extra `out` here a typo?

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -778,4 +778,38 @@ DLL_EXPORT int get_liq_liq_shear_force_per_volume_input_variable(
...
+    @param[in] mu_disp Dispersed Field Viscosity [Pa.s].
+    @param[in] mu_cont Continuous Field Viscosity [Pa.s].
+    @param[in] alpha_disp_in_layer   Dispersed Field Volume Fraction in the layer (emulsion) [m3 of dispersed field /m3 of layer].
```
arthursoprana, line 796:
```suggestion
    @param[in] alpha_disp_in_layer Dispersed Field Volume Fraction in the layer (emulsion) [m3 of dispersed field /m3 of layer].
```

I'm not sure how the indentation works here, but the spacing is different, is it supposed to be like that?

arthursoprana (commented):
Nice work!

I've just added some suggestion related to docs. Please review them as I'm not 100% if they all make sense ;)

docs/source/plugins/06_solver_hooks.rst
ggrbill: Yes, exactly. this is the idea.
I will rephrase it. It is a little confusing.

docs/source/plugins/06_solver_hooks.rst
ggrbill: Nice suggestion.
Thanks.

src/alfasim_sdk/alfasim_sdk_api/api.h
ggrbill: Yes.

- [ ] remove extra `out`

PR #176, opened 2021-06-09
Fb asim 4177 implement pid controller calc

edersongrein:
Support for the implementation of controllers


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1319,6 +1319,76 @@ def _validate_liquid_separation_efficiency(self, attribute, value):
...
+    max_value: float = attr.ib(default=1.0e50, converter=float)
+    max_rate_of_change: float = attr.ib(default=1.0e50, converter=float)
+
+    @max_rate_of_change.validator
+    def _validate_max_rate_of_change(self, attribute, value):
+        assert isinstance(value, float) and value >= 0.0
+
+
+@attr.s(slots=True, kw_only=True)
+class ControllerNodePropertiesDescription:
```
ggrbill, line 1361: I think it makes more sense to be an Equipment. However, currently, we have the `EquipmentDescription` inside the `PipeDescription`, which doesn't make sense to the controller concept, since `InputSignal` and `OutputSignal` could be from different Pipes and Equipment.

However, it bothers me a little bit about the insertion of a controller as a Node, as we already discussed a little bit in the stash PR, but in that PR it could make sense due to our core structure and it would be exposed to the final user. However, in the Case Description semantics, the final user must know everything and I think it would be strange to add a new controller as a `NodeDescription` inside the `nodes` list.

I would suggest creating a new list called `controllers` (just like `pipes` and `nodes` inside the `CaseDescription`) and renaming `ControllerNodePropertiesDescription` to just `ControllerDescription`.

Since it is something that could be only a particular opinion, I would like to hear the @arthursoprana, @tarcisiofischer, and @prusse-martin opinions. What do you guys think?

ggrbill (changes requested):

src/alfasim_sdk/_internal/alfacase/case_description.py
ggrbill: I think @prusse-martin is on vacation. So, @williamjamir's opinion would be also very helpful.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1319,6 +1319,76 @@ def _validate_liquid_separation_efficiency(self, attribute, value):
         assert isinstance(value, Scalar) and 0.6 <= value.GetValue("-") <= 1.0
 
 
+@attr.s(slots=True, kw_only=True)
+class ControllerInputSignalPropertiesDescription:
```
capmayer, line 1323: Hey, nice work!

As it still fresh on your mind, could you please add some docs here?

edersongrein: Sure

src/alfasim_sdk/_internal/alfacase/case_description.py
edersongrein: I will try to refactorate code to not use NodeDescription and see how it goes

edersongrein: I decided to keep the implementation as it is and let whoever pick task ASIM-4178 to decide whether Controller should be a node or not

ggrbill: Ok. No problem.

ggrbill (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
prusse-martin: I agree with @ggrbill that being represented as a node is an implementation detail in alfasim an that should no leak to case description.
Wells are also internally represented by nodes in alfasim but listed separately from edges and "normal" nodes.

edersongrein: Ok, no problem. I will tackle it on ASIM-4255

PR #186, opened 2021-09-24
Fb asim 4372 internal leak

edersongrein:
Support for leaks with orifice model


ggrbill:
Did you remember which files did you really changed?
Because there are the generation of files, I could remember which one (or ones) we change to add a new `description`.

edersongrein:
> Do you remember which files did you really changed?
> Because there is the generation of many files, I couldn't remember which one (or ones) we should change to add a new `description`.

Take a look on file case_description.py. The reset are generated modifications

ggrbill:
> > Do you remember which files did you really changed?
> > Because there is the generation of many files, I couldn't remember which one (or ones) we should change to add a new `description`.
> 
> Take a look on file case_description.py. The reset are generated modifications

Ok. Thanks!

ggrbill (approved):

arthursoprana (approved):

PR #188, opened 2021-10-01
Fb asim 4397 external leak

edersongrein:
Support for external leak (leak from pipe to environment


ggrbill (approved):

ggrbill:
Do not forget to Update the Docs. I believe it is just to add a new entry for `LeakType` enum class.

PR #189, opened 2021-10-04
Fb etk 1692 coilib50 mypy

nicoddemus:


nicoddemus:
Failure is unrelated, I will investigate it afterwards.

oliveira-mauricio (approved):

PR #200, opened 2021-10-14
Implement leak equipment in gui

dettmerramon:


dettmerramon:
long story short, I changed the name of "target_pipe" variables to "target_pipe_name". I only realized later that it broke some tests :v (was planning to let it slide as "target_pipe" but @oliveira-mauricio asked me to change it, so I left the PR open)

oliveira-mauricio:
> long story short, I changed the name of "target_pipe" variables to "target_pipe_name". I only realized later that it broke some tests :v (was planning to let it slide as "target_pipe" but @oliveira-mauricio asked me to change it, so I left the PR open)

Just run `inv cog` locally to update some files, also run the tests and see what you need to update (some times you need to update the expected for some tests).

PR #204, opened 2021-10-29
Add documentation to PIG related enums ASIM-4380

oliveira-mauricio:
Forgot to add this on the merged PR, just a minor update of PIG enums docs.

### DIFF
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -398,28 +398,28 @@ class ControllerType(Enum):
 class PigTrappingMode(Enum):
     """
     Trapping mode of a PIG equipment.
-
-    - Automatic: the Pig is automatically trapped whenever it reaches a boundary node (e.g.,
-        pressure or mass boundary).
-    - UserDefined: the Pig is trapped according to user-defined pipe and position unless it
-        reaches a boundary node first.
     """
 
+    #: The PIG is automatically trapped whenever it reaches a boundary node (e.g.,
+    #: pressure or mass boundary).
     Automatic = "automatic"
+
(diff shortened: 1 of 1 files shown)


src/alfasim_sdk/_internal/constants.py
```diff
@@ -396,11 +396,29 @@ class ControllerType(Enum):
 
 
 class PigTrappingMode(Enum):
+    """
+    Trapping mode of a PIG equipment.
+
+    - Automatic: the Pig is automatically trapped whenever it reaches a boundary node (e.g.,
```
nicoddemus, line 402: We can actually document each enum value using a special comment above it:

```python
class PigTrappingMode(Enum):
    """
    Trapping mode of a PIG equipment.    
    """

    #: The Pig is automatically trapped whenever it reaches a boundary node (e.g.,
    #: pressure or mass boundary).
    Automatic = "automatic"

    #: The Pig is trapped according to user-defined pipe and position unless it
    #: reaches a boundary node first.
    UserDefined = "user_defined"
```

This works for every type of class definition, not being specific to enum.

This renders it like this:

```
![image](https://user-images.githubusercontent.com/1085180/139660737-1e935124-3d1d-4202-8398-db6313be0a9f.png)
```

nicoddemus: I changed those locally to test them, decided to just push the changes. 👍

nicoddemus (approved):

nicoddemus (approved):

src/alfasim_sdk/_internal/constants.py
oliveira-mauricio: That is nice, didn't know about that.

prusse-martin (approved):

PR #209, opened 2021-11-25
Remove "force per square velocity", definition in barril

prusse-martin:
ASIM-4452


prusse-martin:
Jobs will be rerun after merging https://github.com/ESSS/barril/pull/74 and releasing barril. When that happens I will also update the pins.

prusse-martin:
This was abandoned in favor of https://github.com/ESSS/alfasim-sdk/pull/214.

PR #211, opened 2021-11-29
(ASIM-4457) rename discharge coeff attributes

dettmerramon:


oliveira-mauricio (approved):

nicoddemus (approved):

prusse-martin (changes requested):
This a breaking change.
We will not be able to load old alfacases containing a gas lift valve.

This is missing a changelog entry.
We really want to add that breaking change?

nicoddemus:
> We really want to add that breaking change?

I would like, for consistency. Better now than never I think.

Don't we have a mechanism to allow loading attributes with different names? Can we estimate if there are many alfacase files out there which use gas lift valves?

Unless I'm mistaken users of alfascase files are ourselves and PWDa, so I suspect the change is OK to apply now.

prusse-martin (approved):

prusse-martin:
> Don't we have a mechanism to allow loading attributes with different names?

No, we don't.

> Unless I'm mistaken users of alfascase files are ourselves and PWDa, so I suspect the change is OK to apply now.

OK...

PR #219, opened 2021-12-29
(ASIM-4481) AnnulusBasedEquipment refactor

dettmerramon:


src/alfasim_sdk/_internal/constants.py
```diff
@@ -378,6 +378,16 @@ class OutputAttachmentLocation(Enum):
 LeakLocation = OutputAttachmentLocation
 
 
+class EquipmentAttachmentLocation(Enum):
+    """
+    Equipment Attachment Location will tell the location in which this equipment can be attached.
```
nicoddemus, line 383: General tip: you don't need to repeat the name of the class/enum you are documenting. Brevity is good. 😁 

```suggestion
    Location in which an equipment can be attached.
```

nicoddemus (approved):

nicoddemus:
Btw make sure to cleanup the commit history before merging. 👍

PR #225, opened 2022-01-20
Add FlowPatternModel enum

arthursoprana:


oliveira-mauricio:
This new property is not being exported/imported to .alfacase, is that intended?

src/alfasim_sdk/__init__.py
```diff
@@ -255,6 +255,7 @@ def get_alfasim_sdk_api_path():
 from alfasim_sdk._internal.constants import PipeThermalModelType
 from alfasim_sdk._internal.constants import PipeThermalPositionInput
 from alfasim_sdk._internal.constants import PumpType
+from alfasim_sdk._internal.constants import PVTCompositionalViscosityModel
```
prusse-martin, line 258: Do you have any idea of how those did sneak in in the wrong order?
Don't we run precommit on ci?

prusse-martin (approved):

src/alfasim_sdk/__init__.py
arthursoprana: Not sure, but I've sorted them in this PR

arthursoprana:
> This new property is not being exported/imported to .alfacase, is that intentional?
> 
> @arthursoprana just to make sure you read this

@oliveira-mauricio I'll add it as a TODO task of ASIM-4574

oliveira-mauricio (approved):

PR #227, opened 2022-02-08
Support flow pattern model and regime capturing mesh treshold in case description

oliveira-mauricio:
New edge association attributes ASIM-4574

### DIFF
--- a/docs/source/alfacase_definitions/PipeDescription.txt
+++ b/docs/source/alfacase_definitions/PipeDescription.txt
@@ -16,6 +16,8 @@
             environment: \ :class:`EnvironmentDescription <EnvironmentDescription>`\  = EnvironmentDescription()
             segments: \ :class:`PipeSegmentsDescription <PipeSegmentsDescription>`\  = PipeSegmentsDescription()
             initial_conditions: \ :class:`InitialConditionsDescription <InitialConditionsDescription>`\  = InitialConditionsDescription()
+            flow_pattern_model: \ :class:`FlowPatternModel <alfasim_sdk._internal.constants.FlowPatternModel>`\  = FlowPatternModel.UnitCell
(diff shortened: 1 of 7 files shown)


nicoddemus (approved):

docs/source/alfacase_definitions/PipeDescription.txt
```diff
@@ -16,6 +16,8 @@
             environment: \ :class:`EnvironmentDescription <EnvironmentDescription>`\  = EnvironmentDescription()
             segments: \ :class:`PipeSegmentsDescription <PipeSegmentsDescription>`\  = PipeSegmentsDescription()
             initial_conditions: \ :class:`InitialConditionsDescription <InitialConditionsDescription>`\  = InitialConditionsDescription()
+            flow_pattern_model: \ :class:`FlowPatternModel <alfasim_sdk._internal.constants.FlowPatternModel>`\  = FlowPatternModel.UnitCell
+            regime_capturing_mesh_threshold: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(0.0, '-', 'dimensionless')
```
prusse-martin, line 20: I realize this will not be used by default since `flow_pattern_model = FlowPatternModel.UnitCell`. But...

Is `0` a good default for this?

prusse-martin (approved):
I left a question =)

docs/source/alfacase_definitions/PipeDescription.txt
arthursoprana: The default equals to zero basically makes the `FlowPatternModel.RegimeCapturing` equals to `FlowPatternModel.UnitCell`. Since we are still validating `RegimeCapturing`, I think we can keep zero and update the default once we find the most suitable value based on validation resutls.

oliveira-mauricio: Nice, we just need to update both core model and here default case description when we find a better default value for it.

PR #228, opened 2022-02-11
Update pre commit config to use python 3.7

prusse-martin:
Update python version used by pre-commit.
Same as: https://github.com/ESSS/qmxgraph/pull/123

In qmx some tests failed due this and the problem was detected before merging, here all tests passed.
@nicoddemus, do you have any idea why?


nicoddemus:
> In qmx some tests failed due this and the problem was detected before merging, here all tests passed.
@nicoddemus, do you have any idea why?

Not from the top of my head. 🤔

arthursoprana (approved):

PR #233, opened 2022-03-14
(ASIM-4551) Add combined pvt to case description

dettmerramon:


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -505,9 +505,9 @@ class BipDescription:
 
 
 @attr.s(frozen=True, slots=True)
-class FluidDescription:
+class CompositionalFluidDescription:
```
prusse-martin, line 508: Renaming classes is an api breakage, alfasim-sdk is on version 0 but I don't think we should get used to this.

dettmerramon: I see, honestly we should try to be more specific during naming so that we don't let something generic like this happen so often, but I guess the only other option now (assuming we couldn't rename) would be to make FluidDescription contain both compositional and combined fluids

prusse-martin: @nicoddemus 
Are you in favor we make that breakage now?
No many ppl are using it and we are in version ´0`.

nicoddemus: Yeah I think this is fine for now, we just need to warn the PWPa folks.

nicoddemus (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
nicoddemus: Make sure to update the https://github.com/ESSS/alfasim-sdk/blob/master/CHANGELOG.rst too, including all breaking changes.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2553,13 +2553,12 @@ class PvtModelsDescription:
...
+    table_based_pvt_models: Dict[str, Union[str, Path]] = attr.ib(
         default=attr.Factory(dict), validator=dict_of((str, Path))
     )
-
-    correlations = attrib_dict_of(PvtModelCorrelationDescription)
-    compositions = attrib_dict_of(PvtModelCompositionalDescription)
-    combined = attrib_dict_of(PvtModelCombinedDescription)
+    correlation_based_pvt_models = attrib_dict_of(PvtModelCorrelationDescription)
+    compositional_based_pvt_models = attrib_dict_of(PvtModelCompositionalDescription)
+    combined_based_pvt_models = attrib_dict_of(PvtModelCombinedDescription)
```
prusse-martin, line 2561: Now it is a bit too long to me...

prusse-martin: Also this will not be an programmatic breakage, but all existing alfacase files using the old members will break.

nicoddemus: I think we should go with @arthursoprana's guts (I particularly prefer longer names that are descriptive than shorter but cryptic). 

About the breakage, we are already breaking other stuff, might as well improve this part.

prusse-martin: For some comparison.

Before this is merged alfacases could contain:
```yaml
pvt_models:
  correlations:
    Correlations 1:
      # ...
```

After:
```yml
pvt_models:
  correlation_based_pvt_models:
    Correlations 1:
      # ...
```

I think the long name is a bit redundant.

nicoddemus: @arthursoprana ?

arthursoprana: I agree with @prusse-martin that the names are too long now. But it's better than before.

Anyways I'll put some options together with the current one so that we can decide:

```
# Option 1)
# pvt_models
table_based_pvt_models: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlation_based_pvt_models = attrib_dict_of(PvtModelCorrelationDescription)
compositional_based_pvt_models = attrib_dict_of(PvtModelCompositionalDescription)
combined_based_pvt_models = attrib_dict_of(PvtModelCombinedDescription)

# Option 2)
# pvt_models
table: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlation: = attrib_dict_of(PvtModelCorrelationDescription)
compositional: = attrib_dict_of(PvtModelCompositionalDescription)
combined: = attrib_dict_of(PvtModelCombinedDescription)

# Option 3)
# pvt_models
table_models: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlation_models: = attrib_dict_of(PvtModelCorrelationDescription)
compositional_models: = attrib_dict_of(PvtModelCompositionalDescription)
combined_models: = attrib_dict_of(PvtModelCombinedDescription)

# Option 4)
# pvt_models
table_pvts: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlation_pvts: = attrib_dict_of(PvtModelCorrelationDescription)
compositional_pvts: = attrib_dict_of(PvtModelCompositionalDescription)
combined_pvts: = attrib_dict_of(PvtModelCombinedDescription)

# Option 5)
# pvt_models
tabs: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
corrs: = attrib_dict_of(PvtModelCorrelationDescription)
comps: = attrib_dict_of(PvtModelCompositionalDescription)
combs: = attrib_dict_of(PvtModelCombinedDescription)
```

Option 1) is the current one (the most explicit but the longest as well). For me **Option 4)** looks ok. What do you guys think?

prusse-martin: If we are going to change (and I am against that) I will go with `3` or `4`.

As I already put forward this will possibly break most of `.alfacase`s in the wild.
I use "possibly" and "most" because I don't have the actual numbers but:
- "possibly": those are 3 of 4 existing input modes, and the 5th one is not even documented and only used internally;
- "most": the example I posted was extracted from an `.alfacase` file from my computer, when looking for the example I was ready to look on multiple files before finding one but was the first, and again "3 of 4 existing input modes";

My suggestion is "give a good and kind of matching name to the new input mode and leave the rest untouched by now".
And wait for a backward compatibility to be in place before renaming those.

I will like to keep at a minimum unnecessary "just manually edit your files to match the current valid format".
And avoid a possible _request_ to rush the implementation of backward compatibility.

dettmerramon: damn, just saw it now after making the changes :s, anyways, why are you against long names tho? I personally like it in this context, since it won't impact in readability but will let it clear what its intent is

arthursoprana: @prusse-martin I'll leave two last alternatives here

```
# Option 6)
tables: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlations: = attrib_dict_of(PvtModelCorrelationDescription)
compositional_pvts: = attrib_dict_of(PvtModelCompositionalDescription)
combined_pvts: = attrib_dict_of(PvtModelCombinedDescription)

# Option 7)
tables: Dict[str, Union[str, Path]] = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlations: = attrib_dict_of(PvtModelCorrelationDescription)
compositional_models: = attrib_dict_of(PvtModelCompositionalDescription)
combined_models: = attrib_dict_of(PvtModelCombinedDescription)
```

These last two options have the benefit of being **non-breaking** for `correlations` and `tables` (which are the most common models). The compositional model is still in experimental mode and is probably not being used too much by users.

What do you think?

BTW: Sorry for extending this thread, we may decide it in a meeting if you want.

nicoddemus: Yeah this is blocking the release and has extended enough already.

I suggest we return to the original names (before @arthursoprana's suggestion) if it is just a matter of reverting commits, and defer renaming things to a later date when we have backward compatibility support.

nicoddemus:
```python
# Option 8)
tables = attr.ib(default=attr.Factory(dict), validator=dict_of((str, Path)))
correlations = attrib_dict_of(PvtModelCorrelationDescription)
compositional = attrib_dict_of(PvtModelCompositionalDescription)
combined = attrib_dict_of(PvtModelCombinedDescription)
```

prusse-martin: Remember to update the chagelog with somethink like

```
* **Breaking Change**: Change in ``PvtModelsDescription`` renamed experimental compositional models holder property from ``compositions`` to ``compositional``;

* Example on how to update previous .alfacase files with compositional models:

  - Before::

      pvt_models:
        compositions:
          Compositional model 1:
            equation_of_state_type: pvt_compositional_peng_robinson
            ...  # Other properties

  - After::

      pvt_models:
        compositional:
          Compositional model 1:
            equation_of_state_type: pvt_compositional_peng_robinson
            ...  # Other properties
```

I remember we agreed on the meeting, but exactly why we are renaming this guy a singular while all other are plurals?

prusse-martin (approved):

PR #234, opened 2022-04-02
(ASIM-3288) Expose black-oil properties 

dettmerramon:


src/alfasim_sdk/__init__.py
```diff
@@ -517,4 +520,7 @@ def get_alfasim_sdk_api_path():
     "load_pvt_model_table_parameters_description_from_alfatable",
     "tab",
     "tabs",
+    "CorrelationsOilViscosity",
```
prusse-martin, line 523: This list should be kept in alphabetical order.

src/alfasim_sdk/_internal/constants.py
```diff
@@ -465,3 +465,15 @@ class FlowPatternModel(Enum):
 
 
 MULTI_INPUT_TYPE_SUFFIX = "_input_type"
+
+
+class CorrelationsOilViscosity(Enum):
```
prusse-martin, line 470: Those enums are not used in case description? (or alfacase?)

src/alfasim_sdk/__init__.py
prusse-martin: Also note as lower case values come after capitalized ones.

src/alfasim_sdk/_internal/constants.py
dettmerramon: not as of now

src/alfasim_sdk/__init__.py
dettmerramon: ok

src/alfasim_sdk/__init__.py
```diff
@@ -351,6 +354,9 @@ def get_alfasim_sdk_api_path():
     "ControllerType",
     "CorrelationPackage",
     "CorrelationPackageType",
+    "CorrelationsOilViscosity",
+    "CorrelationsGasViscosity",
```
prusse-martin, line 358:
```suggestion
    "CorrelationsGasViscosity",
    "CorrelationsOilViscosity",
```

prusse-martin (approved):

PR #236, opened 2022-04-06
(ASIM-4648) Add alfacase support for black-oil properties

dettmerramon:


prusse-martin:
There are changes that will affect generated code (the schema and documentaton).

prusse-martin (approved):

nicoddemus (changes requested):
LGTM, please update the CHANGELOG with the new properties. 👍

prusse-martin:
Did you change defaults.
Why?

dettmerramon:
> You did change defaults. Why?

Synthetic tests were failing because of it

PR #247, opened 2022-05-18
Fb asim 4672 missing emulsion models

edersongrein:


CHANGELOG.rst
```diff
@@ -42,6 +42,20 @@ History
...
+  - ``pal_rhodes1989``;
+  - ``ronningsen1995``;
+  - ``volumetric_weight``;
+  - ``woelflin_1942``;
+  - ``barnea_mizrahi1976``;
+  - ``table_based``;
+
+* Add emulsion constant inversion point model;
+
+* Add emulsion relative viscosity tuning factor;
```
ggrbill, line 58: Nice! 
I forgot to ask you to add this entry in the previous PR.

src/alfasim_sdk/_internal/constants.py
```diff
@@ -115,6 +115,12 @@ class EmulsionRelativeViscosityModelType(Enum):
     Mooney1951a = "mooney1951a"
     Mooney1951b = "mooney1951b"
     FromPlugin = "from_plugin"
```
ggrbill, line 117: I prefer to put `FromPlugin` option as the last one. 
But I am not sure if it will break something with the `Enum`.

ggrbill (approved):

src/alfasim_sdk/_internal/constants.py
edersongrein: I think this would require some treatment when loading old cases. I don't know if it is worthy to do such modification

prusse-martin: We save the "string" not some auto index or some thing like that.
That being said, adding new values (any where in the enum) will not affect loading old cases.
Changing values (the strings) or removing the entries will affect old cases using the old values.

ggrbill: Nice! Thanks for the explanation @prusse-martin!

src/alfasim_sdk/_internal/hook_specs.py
```diff
@@ -1785,6 +1785,8 @@ def calculate_relative_emulsion_viscosity(
     mu_disp: "double",
     mu_cont: "double",
     alpha_disp_in_layer: "double",
+    T: "double",
```
prusse-martin, line 1788: I think that changing hook signatures will cause incompatibility with existing plugins.
And they will need to be updated (source code changed, not just recompiled).
Changelog entry about the breaking change?
Or we could go with creating an alternative version for the hook like `calculate_relative_emulsion_viscosity_per_temperature` or some think else in those lines (does `T` stands for temperature?).

I also think you should update the docstring to properly document the signature, I think that is what goes to the online documentation.

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -908,14 +908,18 @@ DLL_EXPORT int get_liq_liq_shear_force_per_volume_input_variable(
...
+    @param[in] water_in_oil True when water is dispersed in oil, otherwise it is a dispersion of oil in water.
     @return An #error_code value.
 */
 DLL_EXPORT int get_relative_emulsion_viscosity(
     void* ctx,
     double* out,
     double mu_disp,
     double mu_cont,
-    double alpha_disp_in_layer
+    double alpha_disp_in_layer,
```
prusse-martin, line 920: Related to {{comment on hook_specs.py}}.

src/alfasim_sdk/_internal/constants.py
prusse-martin: But I also prefer `FromPlugin` kind placed apart from the others, like last or first.
Having it as first has the benefit of getting out of change sets, I will put as a possible problem with that that it will be the default value, but on case description no default is assumed for enums.
That will probably be an issue with alfasim, where every thing has defaults and possibly some enum use implicit defaults (the "first value") but that can be fixed without backward compatibility effects and easily detectable checking the status.

prusse-martin (changes requested):

src/alfasim_sdk/_internal/constants.py
edersongrein: You are right @prusse-martin. 

- [ ] Replace plugin entry

src/alfasim_sdk/_internal/hook_specs.py
edersongrein: This will break the existing plugins indeed and I forgot to update changelog and the docstring.

- [ ] Update changelog and docstring

> Or we could go with creating an alternative version for the hook like calculate_relative_emulsion_viscosity_per_temperature or some thing else in those lines (does T stands for temperature?

Yes, T is for temperature and creating another hook is not viable. Fortunately as far as I know there is nobody using this plugin, so this change probably won't bother anyone

edersongrein: In fact I have updated the changelog. It was only the docstring that was missed

prusse-martin: > ... creating another hook is not viable ... there is nobody using this plugin...

OK... I will let it pass for now =P

prusse-martin (approved):

PR #251, opened 2022-06-15
Reviewed esp speed unit

arthursoprana:
Reviewed units for speed, using only Hz, rad/s and rpm and fix esp_speed category

### DIFF
--- a/src/alfasim_sdk/_internal/units.py
+++ b/src/alfasim_sdk/_internal/units.py
@@ -158,3 +158,11 @@ def register_units() -> None:
         valid_units=std_volume_units,
     )
     # NOTE: "gas volume flow rate" and "gas volume" are also required, but they are already registered
+
+    angle_per_time_units = ["Hz", "rpm", "rad/s"]
+    db.AddCategory(
+        "angle per time",
+        quantity_type="frequency",
+        valid_units=angle_per_time_units,
+        override=True,
+    )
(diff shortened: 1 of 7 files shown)


src/alfasim_sdk/_internal/units.py
```diff
@@ -158,3 +158,11 @@ def add_category_if_not_defined(*, category, quantity_type, valid_units):
         valid_units=std_volume_units,
     )
     # NOTE: "gas volume flow rate" and "gas volume" are also required, but they are already registered
+
+    angle_per_time_units = ["Hz", "rpm", "rad/s"]
+    db.AddCategory(
+        "angle per time",
```
ggrbill, line 164: Is this to simplify conversions between those units?
I am asking because Frequency and Angular Velocity are a little bit different, even though they are interchangeable.

ggrbill (approved):

src/alfasim_sdk/_internal/units.py
arthursoprana: Here I'm simply restricting the available units for category `angle per time`. With this change, we'll only show `["Hz", "rpm", "rad/s"]` to the user.

> I am asking because Frequency and Angular Velocity are a little bit different, even though they are interchangeable.

I'm guessing that this question is related to that fact that you saw `frequency` somewhere in the diff. That was incorrect, it should have been `category="angle per time"` and not `category="frequency"`.

ggrbill: Actually, I was asking because usually `rad/s` is considered an angular velocity(`angle per time`), and `Hz` is considered a frequency (`inverse of time` or `1 per time`). We can covert one to another, but they are not the same.

I think you answered my question, it was made to simplify. Since they are interchangeable, we can consider the same category and make the conversions implicitly.

BeneBr (approved):

src/alfasim_sdk/_internal/units.py
prusse-martin: > Here I'm simply restricting the available units...

Maybe we should add this to change log.

PR #252, opened 2022-06-17
Add missing automatic_{profile,trend}_frequency parameter

cauebs:


src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
```diff
@@ -1592,8 +1592,10 @@ def load_case_output_description(
 ) -> case_description.CaseOutputDescription:
```
cauebs, line 1592: Should I add a unit test for this function somewhere?

ggrbill: I believe it is already tested since you just added those new parameters to the `filled_case_description`.

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -441,8 +441,10 @@
 )
 CASE_OUTPUT_DESCRIPTION = case_description.CaseOutputDescription(
     trends=TRENDS_OUTPUT_DESCRIPTION,
+    automatic_trend_frequency=True,
```
ggrbill, line 444: What is the idea of these parameters? If they are `True`, the `trend_frequency` and `profile_frequency` are ignored?

ggrbill (changes requested):
Please, add a new entry to `CHANGELOG.rst`.

src/alfasim_sdk/_internal/alfacase/alfacase_to_case.py
cauebs: You'd think that, but the tests pass regardless of the two fields added to this dict.

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
cauebs: They already exist, but are missing in some places. If I understand them correctly, it's like what you said, `automatic_...` being `True` means the property will be derived automatically from some defaults.

ggrbill (approved):

nicoddemus (approved):

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
prusse-martin: It is also possible that the "case description"-"alfacase" round trip is saving the unused values. I remember to add that option with the multi inputs to test transient and constant in one go (but disabled on production).

But as he pointed out "... used is the highest value between: default profile frequency, maximum timestep configured, and an estimate ..."

PR #259, opened 2022-07-25
Initial alfasim_runner fixture

prusse-martin:
HYD-431


nicoddemus (approved):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+            # Take note of any file to ignore.
+            name, _ = os.path.splitext(f)
+            files_under_creation.add(Path(name))
+        elif ext == "":
+            result_files.add(f)
+        else:
+            raise ValueError(f"Unknown ext: {ext}")
+
+    prefix_len = len(RESULT_FILE_PREFIX)
+    result_files.difference_update(files_under_creation)  # Ignore incomplete files.
```
nicoddemus, line 207: IIUC, this has no effect, because you never add an incomplete file to `result_files`?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+                f"the simulation.\n"
+                "\n"
+                f'1 - Click in "Save as".\n'
+                f"2 - Select a different path, outside Download folder.\n"
+                f"3 - Run simulation again.\n"
+            )
+        finally:
+            os.chdir(old_directory)
+
+    # When a new result file is created its metadata contents are not complete, and the file
```
nicoddemus, line 181: A safer approach would be to actually write the result files to a temporary file name (in the same directory), using an extension which wouldn't be globbed here, and only when the file is complete, you move the file to the final name using `os.replace`.

You can then simplify this code because now there's no special handling that needs to be done.

Or am I missing something?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+    result_files_sorted = (
+        (int(filename.name[prefix_len:]), open_result_file(filename))
+        for filename in result_files
+    )
+    result_files_sorted_dict = _Dict(sorted(result_files_sorted, key=lambda x: x[0]))
+
+    # Close the files once the container is collected.
+    import weakref
+
+    weakref.finalize(
```
nicoddemus, line 217: I see what you did here, but perhaps it would be better to make this a context manager, which closes the files at the end.

Using weakrefs for this is a bit implicit/magical, so I would prefer a more explicit approach if possible.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+                size=next_old_info.size,
+                uuid=next_new_uuid,
+            )
+
+    return time_set_info
+
+
+def _global_index_to_file_based_index(
+    index: int, time_set_start: int, time_set_size: int
+):
```
nicoddemus, line 500:
```suggestion
) -> int:
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+        """
+        Read a metadata file and add it to a metadata collection.
+
+        :param metadata_collection:
+            This collection is changed. The keys are the "base time step index"
+            and the values are the metadata from the file (the duplicates have
+            been removed and the original index recorded).
+        """
+        meta = json.loads(result_file[META_GROUP_NAME].attrs[output_type])
+        # # Remove duplicates from the result. Save the original index.
```
nicoddemus, line 780: Commented out code?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1344 @@
...
+def _read_profile_arrays(
+    result_metadata,
+    output_keys,
+    time_step_index,
+    *,
+    group_name: str,
+    data_attr: str,
+    data_id_suffix: str = "",
+    slicer: Callable[[int], Any],
+):
```
nicoddemus, line 848: Please add return type annotation

src/alfasim_sdk/result_reader/reader.py
```diff
@@ -0,0 +1,252 @@
...
+from alfasim_sdk.result_reader.aggregator import read_profiles_data
+from alfasim_sdk.result_reader.aggregator import read_profiles_domain_data
+from alfasim_sdk.result_reader.aggregator import read_time_sets
+from alfasim_sdk.result_reader.aggregator import read_trends_data
+from alfasim_sdk.result_reader.aggregator_constants import RESULTS_FOLDER_NAME
+
+
+class Results:
+    """
+    This class abstracts read trend and profile curves
```
nicoddemus, line 22:
```suggestion
    Allows reading trend and profile curves
```

nicoddemus: Btw I might have missed this, but does this class appear in the documentation?

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -82,6 +97,45 @@ class PluginTableContainer:
     columns = attr.ib(default=attr.Factory(dict))
 
 
+@attr.s(frozen=True, slots=True)
+class PluginFileContent:
+    path = attr.ib(validator=instance_of(Path))
+    content = attr.ib(validator=instance_of(bytes))
+    size = attr.ib(validator=instance_of(int))
+    modified_data = attr.ib(validator=instance_of(datetime))
```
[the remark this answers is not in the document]
prusse-martin: I will take the chance to fix this typo in ben10 and update the sdk documentation.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -82,6 +97,45 @@ class PluginTableContainer:
...
+    is_valid = attr.ib(validator=instance_of(bool))
+
+    @classmethod
+    def from_path(cls, file_path: Path, root: Optional[Path] = None):
+        """
+        Helper method for creation of FileContent from a Path object or a string with the absolute path.
```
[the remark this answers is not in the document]
prusse-martin: I will update the docs.
Basically allow `file_path` to be an absolute path and the `path` member of the resulting object to be a relative value.

src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
```diff
@@ -0,0 +1,404 @@
...
+
+def get_unused_filename(alfacase_folder: Path, path: Path) -> Path:
+    """
+    Return a not used file name inside ``alfacase_folder``.
+    """
+    n = 1
+    extra = ""
+    while True:
+        fullpath = alfacase_folder / f"{path.stem}{extra}{path.suffix}"
+        if not fullpath.exists():
```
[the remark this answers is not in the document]
prusse-martin: I want to test that id do not exits (not a folder and not an file).
For negative tests I think the more generic `not path.exists()` a better choices or I will need to do `if not path.is_file() and not path.is_dir()``.

src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
```diff
@@ -0,0 +1,404 @@
...
+        return value
+    raise InvalidPluginDataError(
+        f"Can not convert to an enum: {value!r}"
+        f" (valid values are {type_from_plugin.values!r}"
+    )
+
+
+def _convert_file_content(
+    value: object, type_from_plugin: BaseField, alfacase_path: Path
+) -> object:
```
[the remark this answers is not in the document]
prusse-martin: I was going for a homogeneous signature.
But `Callable[..., object]` will happily receive a `Callable[..., PluginFileContent]`.
I will update those.

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin: On Windows when moving the file the writer process need to close the file.
In the time window between "move" and "reopen the file in swmr write mode" a reader could try to open the file in swmr reading mode, this will fail (a swmr writter must open the file first) and the open in normal mode fallback will be used, causing inconsistencies when the writer process reopen the file.
I actually don't remember if the "open for swmr write" will fail or not (open to read when no writer is presente raises).

I welcome suggestions to improve the comment =)

src/alfasim_sdk/_internal/alfacase/case_description.py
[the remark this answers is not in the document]
nicoddemus: ben10?

src/alfasim_sdk/_internal/alfacase/case_description.py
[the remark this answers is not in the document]
nicoddemus: Perhaps we can make that less error prone by checking `file_path.is_absolute()` if `root` is given?

src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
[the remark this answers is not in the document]
nicoddemus: Hmm but why do you care if a folder with same name exists, if you are trying to create a file?

But no problem, just pointing this out because I've seen code which wanted to check for files only, but still used `exists`... but if you really care about both, leave is as. 👍

src/alfasim_sdk/result_reader/aggregator.py
nicoddemus: Not sure, that was just a normal comment... but are you aware that `os.replace` is atomic (contrary to `os.rename` which is not)?

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin:
```
0
0.creating
```
0 will be added to `result_files` because "0" is present in the glob and later 0 will be added to `files_under_creation` due "0.creating".

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin: That was a bad decision when I wrote it...
I will check this.

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin: I will remove this.
It was already commented in alfasim for a long time =)

src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
```diff
@@ -495,3 +495,9 @@ class InvalidYamlDataError(DescriptionError):
     """
     Error raised when some data in the YAML file is not properly configured.
     """
+
+
+class InvalidPluginDataError(DescriptionError):
+    """
+    Error raised when some plugin is not in the expected format.
```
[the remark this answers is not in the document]
prusse-martin: I will change to
```
"""
Error raised when some plugin input data is not in the expected format.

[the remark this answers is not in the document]
Comment (nicoddemus) [reply src/alfasim_sdk/_internal/alfacase/case_description_attributes.py]: Perfect!

File: src/alfasim_sdk/result_reader/aggregator.py (hunk 2 above)
Comment (nicoddemus) [reply src/alfasim_sdk/result_reader/aggregator.py:187]: Why can't you `put hdf in swmr mode` in the first case before closing? Sorry I don't remember the swmr details.

Comment (nicoddemus) [reply src/alfasim_sdk/result_reader/aggregator.py:187]: Or you can only `put hdf in swmr mode` for the file if it is in its final filename?

Comment (prusse-martin) [reply src/alfasim_sdk/result_reader/aggregator.py:187]: The swmr works by the writer opening the file first, when closing the file is no longer open.

Comment (nicoddemus) [reply src/alfasim_sdk/result_reader/aggregator.py:187]: Ahh got it. 👍 Thanks for the explanation.

File: src/alfasim_sdk/result_reader/aggregator.py (hunk 3 above)
Comment (nicoddemus) [reply src/alfasim_sdk/result_reader/aggregator.py:247]: Gentle ping here

**Review by nicoddemus — APPROVED** [review]

File: src/alfasim_sdk/result_reader/aggregator.py [hunk 13]
```diff
@@ -0,0 +1,1430 @@
...
+        List of sourced time set keys (tuples of relevant base time sets).
+    """
+
+    app_version_info = attr.ib(validator=attr.validators.instance_of(dict))
+    profile_key_to_metadata = attr.ib(validator=attr.validators.instance_of(dict))
+    trend_key_to_metadata = attr.ib(validator=attr.validators.instance_of(dict))
+    time_set_keys = attr.ib(validator=attr.validators.instance_of(list))
+
+
+# load_result_files
```
nicoddemus, line 173:
```suggestion
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -0,0 +1,1430 @@
...
+@contextmanager
+def open_result_files(result_directory: Path) -> Dict[int, h5py.File]:
```
nicoddemus, line 175: Nice solution. 👍

BeneBr (approved):

dettmerramon (approved):

cauebs (approved):

src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
```diff
@@ -0,0 +1,402 @@
...
+    """
+    n = 1
+    extra = ""
+    while True:
+        fullpath = alfacase_folder / f"{path.stem}{extra}{path.suffix}"
+        if not fullpath.exists():
+            return fullpath
+        else:
+            n += 1
+            extra = f" ({n})"
```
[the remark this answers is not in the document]
prusse-martin:
```
 PurePath.with_stem(stem)
    ...
    New in version 3.9.
```
=/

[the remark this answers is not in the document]
prusse-martin:
```python

    # The original file does not need to actually exist (`path.stem` could be empty).
    stem = path.stem or '(unnamed)'
    fullpath = alfacase_folder / f"{stem}{path.suffix}"
    for i in itertools.count(2):
        if fullpath.exists():
            fullpath = fullpath.with_name(f"{stem} ({i}){path.suffix}")
        else:
            return fullpath
```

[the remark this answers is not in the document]
nicoddemus: Avoid `exists()`, prefer to be more explicit with `is_dir()` and `is_file()`.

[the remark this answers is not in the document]
prusse-martin: No `exists`is the correct one here.
If there is a folder with that name I will be unable to create a file with the same name.
Or are you saying instead of `if x.exists():` I should do `if x.is_dir() or x.is_file():`, not considering things like block devices,. sockets and stuff?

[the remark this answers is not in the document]
nicoddemus: btw why `itertools.count(2)`?

PR #261, opened 2022-08-03
Fb asim 4686 calculate temperature from enthalpy

Gualcar:
This is a tiny modification inside the `PvtModelTableParametersDescription`.

A new parameter `table_type` have been created to indicate if the table is specified in terms of PT or PH. This is important in this file `alfasim_core/source/python/alfasim_core/pvt_model/__init__.py` where we initialize the `PvtModelTableParametersDescription`.

Also, this function `build_linear_initial_pressure_description` have been added to easily create a linear pressure profile for the initial condition.


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2243,6 +2243,10 @@ class PvtModelTableParametersDescription:
     number_of_phases: int = attr.ib(default=2, validator=instance_of(int))
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
+    table_type: Optional[str] = attr.ib(
+        default="PT", validator=optional(instance_of(str))
```
arthursoprana, line 2247: We should be using an `Enum` here. Take a look in [here](https://github.com/ESSS/alfasim-sdk/blob/master/src/alfasim_sdk/_internal/constants.py) in order to see how we use it. Using strings for representing `_type` can cause problems.

arthursoprana: Also, can `table_type` be `None`? The `Optional[str]` assumes it can.

ggrbill (approved):
Please, take a look at the @arthursoprana comments.

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: I created a new `Enum` called `PVTTableType`. Their values are the same as defined in `core_constants` to use the `alfasim-sdk` definition to build the `PvtModel` in the `alfasim_core`.

Under the `PvtModelTableParametersDescription` I defined the `table_type` as an `attrib_enum` referencing the created `Enum`.

Now I am using this `Enum` in `ALFAsim` repo  to create the `PvtModel` in `pvt_model.py`

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2243,6 +2243,8 @@ class PvtModelTableParametersDescription:
     number_of_phases: int = attr.ib(default=2, validator=instance_of(int))
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
+    table_type = attrib_enum(default=constants.PVTTableType.PT)
```
arthursoprana, line 2246: 👍

arthursoprana (approved):

PR #264, opened 2022-09-12
ASIM-4933: Implement tests with constant PH tables

Gualcar:
The idea here is to create a constant PH table that is able to test some network equipments, such valves, pump and compressor. The results can be compared directly with the PT model as the same equations have been used.

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/case_description.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description.py
@@ -2522,7 +2522,7 @@ class PvtModelTableParametersDescription:
 
         def temperature(p, h):
             h_lg = 2.260e6
-            return (h - h_lg - h_l_ref)/cp_g_ref
+            return (h - h_lg - h_l_ref) / cp_g_ref
 
         def ideal_gas_density_model(p, t):
             """
@@ -2538,7 +2538,7 @@ class PvtModelTableParametersDescription:
             return 1 / (r * t)
 
         def gas_density_derivative_respect_temperature(p, t):
-            return -p / (r * t ** 2)
+            return -p / (r * t**2)
 
         def constant_density_model(p, t):
             return rho_l_ref + 0 * p
@@ -2705,7 +2705,7 @@ class PvtModelTableParametersDescription:
             water_cut=Scalar(0, "-"),
             total_water_fraction=Scalar(0, "-"),
             number_of_phases=3 if has_water else 2,
-            table_type=constants.PVTTableType.PH.value
+            table_type=constants.PVTTableType.PH.value,
         )
 
     @staticmethod
--- a/src/alfasim_sdk/_internal/alfacase/case_description.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description.py
@@ -2244,7 +2244,7 @@ class PvtModelTableParametersDescription:
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
     table_type: Optional[str] = attr.ib(
-        default=None, validator=optional(instance_of(str))
+        default=constants.PVTTableType.PT.value, validator=optional(instance_of(str))
     )
 
     def __attrs_post_init__(self):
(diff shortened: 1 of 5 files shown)


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2491,6 +2491,195 @@ def oil_water_surface_tension_model(p, t):
...
+        has_water=False,
+    ):
+        """
+        Returns parameters that can be used to create a very simple constant and isothermal PVT
+        table.
+        Used by PvtTable to build it's default table if no parameter is passed.
+        """
+        r = 286.9  # Air individual gas constant [J/kg K]
+
+        def temperature(p, h):
```
Gualcar, line 2523: Temperature as function of enthalpy. Same equation as the `create_constant` method, but here the temperature is in evidence.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2491,6 +2491,195 @@ def oil_water_surface_tension_model(p, t):
...
+            oil_viscosity_model(p, t).flatten(),
+            surface_tension_model(p, t).flatten(),
+            gas_mass_fraction_model(p, t).flatten(),
+            gas_specific_heat_model(p, t).flatten(),
+            oil_specific_heat_model(p, t).flatten(),
+            gas_specific_enthalpy_model(p, t).flatten(),
+            oil_specific_enthalpy_model(p, t).flatten(),
+            gas_thermal_conductivity_model(p, t).flatten(),
+            oil_thermal_conductivity_model(p, t).flatten(),
+            temperature(p,h).flatten()
```
Gualcar: Temperature as an output variable in the last position, in conformity with the PH table style.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2491,6 +2491,195 @@ def oil_water_surface_tension_model(p, t):
...
+            temperature_std=Scalar(15, "degC"),
+            gas_density_std=Scalar(rho_g_ref, "kg/m3"),
+            oil_density_std=Scalar(rho_l_ref, "kg/m3"),
+            water_density_std=Scalar(rho_w_ref, "kg/m3"),
+            gas_oil_ratio=Scalar(0, "sm3/sm3"),
+            gas_liquid_ratio=Scalar(0, "sm3/sm3"),
+            water_cut=Scalar(0, "-"),
+            total_water_fraction=Scalar(0, "-"),
+            number_of_phases=3 if has_water else 2,
+            table_type=constants.PVTTableType.PH.value
```
Gualcar: Here the PH table type is specified and the `mode` of the `PvtModel` can be specified.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2244,7 +2244,7 @@ class PvtModelTableParametersDescription:
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
     table_type: Optional[str] = attr.ib(
-        default=None, validator=optional(instance_of(str))
+        default=constants.PVTTableType.PT.value, validator=optional(instance_of(str))
```
ggrbill, line 2247: Is this `.value` required?

couldn't it be just `default=constants.PVTTableType.PT, ...`?

ggrbill (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: I could try to change it to be an `Enum`. In the first `enthalpy` PR I tried to do it but I had some problems with the type checking for the `Enum`.

Also, this `.table_type` attribute is being passed to the `.mode` attribute that needs to be a `str`. Perhaps the `PvtModel` `mode` attribute could also be changed to an `Enum`. 

This change will impact in other places, since now this is being used also to import the PH tables. I could open a PR only to it if you all agree?

ggrbill: No rush. perhaps create a task to tackle this and do it when you think it is more appropriate.

Gualcar: I added this point into the task [ASIM-4887](https://eden.esss.co/jira/browse/ASIM-4887) description. So I could remember to do it latter.

prusse-martin: Did you use `attrib_enum` similar to this example?
https://github.com/ESSS/alfasim-sdk/blob/565119cbbfbad72e71a34bac5dbcc6ed91ed19d3/src/alfasim_sdk/_internal/alfacase/case_description.py#L94-L97

tests/common_testing/alfasim_sdk_common_testing/case_builders.py
```diff
@@ -147,3 +147,21 @@ def build_compressor_pressure_table_description(
...
+    ):
+        return case_description.PvtModelTableParametersDescription.create_constant()
+    elif energy_model_primary_variable == constants.EnergyModelPrimaryVariable.Enthalpy:
+        return (
+            case_description.PvtModelTableParametersDescription.create_constant_ph_table()
+        )
+    else:  # pragma no cover
+        assert isinstance(
+            energy_model_primary_variable, constants.EnergyModelPrimaryVariable
+        ), f"{energy_model_primary_variable} not supported."
```
prusse-martin, line 167: If we add a new value to `constants.EnergyModelPrimaryVariable` This will not trigger an error even if there is we forget to add code for handling it, a better approach will be just fail:
```suggestion
        assert False, f"{energy_model_primary_variable!r} not supported"
```

src/alfasim_sdk/_internal/alfacase/case_description.py
ggrbill: @Gualcar could you check if @prusse-martin's suggestion solves the problem? I

Gualcar: Thank you for the tip @prusse-martin, I tried using this `attrib_enum` and I had problems with type checking but I not investigated it a lot.

Probably I used it wrong, because there is a difference passing `constants.PVTTableType.PT` (type `PVTTableType`), `constants.PVTTableType` (type `EnumMeta`) or `constants.PVTTableType.PT.value` (type `str`).

I will try it.

tests/alfacase/test_case_description.py
```diff
@@ -888,6 +888,45 @@ def test_pvt_model_table_parameters_description_create_constant():
     np.array_equal(pvt.table_variables[1], expected_value.flatten())
 
 
+def test_pvt_model_table_parameters_description_create_constant_ph_table():
+    import numpy as np
+
+    pressure_values = np.linspace(0.5, 1e10, 4)
+    enthalpy_values = np.linspace(2617360.0, 2869860, 30)
+    h, p = np.meshgrid(enthalpy_values, pressure_values)
+    r = 286.9  # Air individual gas constant [J/kg K]
```
arthursoprana, line 897:
```suggestion
    R = 286.9  # Air individual gas constant [J/kg.K]
```

tests/alfacase/test_case_description.py
```diff
@@ -888,6 +888,45 @@ def test_pvt_model_table_parameters_description_create_constant():
...
+
+    pvt = case_description.PvtModelTableParametersDescription.create_constant_ph_table(
+        ideal_gas=False,
+        rho_g_ref=rho_g_ref,
+        cp_g_ref=cp_g_ref,
+        h_l_ref=h_l_ref,
+        has_water=True,
+    )
+
+    t = pvt.table_variables[-1]
```
arthursoprana, line 914: Does `t` stands for temperature? Perhaps use

```suggestion
    T = pvt.table_variables[-1] # [K]
```

is the unit in `K` or `degC`?

tests/alfacase/test_case_description.py
```diff
@@ -888,6 +888,45 @@ def test_pvt_model_table_parameters_description_create_constant():
...
+
+    # Check gas_density_derivative_respect_pressure
+    gas_density_derivative_expected_value = 1 / (r * t)
```
arthursoprana, line 917:
```suggestion
    gas_density_derivative_expected_values = 1 / (R * T.flatten())
```

Then the `.flatten()` below could be removed.

Also note the change from `_value` to `_values` since it is an array, right?

tests/alfacase/test_case_description.py
```diff
@@ -888,6 +888,45 @@ def test_pvt_model_table_parameters_description_create_constant():
...
+
+    h_lg = 2.260e6
+    temperature_expected_value = (h - h_lg - h_l_ref) / cp_g_ref
+
+    assert np.array_equal(pvt.table_variables[0], gas_density_expected_value.flatten())
```
arthursoprana, line 922: Consider using `pytest.approx` here:

```suggestion
    from pytest import approx
    assert pvt.table_variables[0] == approx(gas_density_expected_values)
```

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2491,6 +2491,226 @@ def oil_water_surface_tension_model(p, t):
...
+            "gas specific enthalpy",
+            "oil specific enthalpy",
+            "gas thermal conductivity",
+            "oil thermal conductivity",
+            "temperature",
+        ]
+
+        if has_water:
+            names.pop()
+            data.pop()
```
arthursoprana, line 2667: In order to avoid these `.pop()` calls, perhaps

```python
data = [
    gas_density_model(p, t).flatten(),
    gas_density_derivative_respect_pressure(p, t).flatten(),
    gas_density_derivative_respect_temperature(p, t).flatten(),
    constant_density_model(p, t).flatten(),
    oil_density_derivative_respect_pressure(p, t).flatten(),
    oil_density_derivative_respect_temperature(p, t).flatten(),
    gas_viscosity_model(p, t).flatten(),
    oil_viscosity_model(p, t).flatten(),
    surface_tension_model(p, t).flatten(),
    gas_mass_fraction_model(p, t).flatten(),
    gas_specific_heat_model(p, t).flatten(),
    oil_specific_heat_model(p, t).flatten(),
    gas_specific_enthalpy_model(p, t).flatten(),
    oil_specific_enthalpy_model(p, t).flatten(),
    gas_thermal_conductivity_model(p, t).flatten(),
    oil_thermal_conductivity_model(p, t).flatten(),
]

names = [
    "gas density",
    "gas density derivative pressure",
    "gas density derivative temperature",
    "oil density",
    "oil density derivative pressure",
    "oil density derivative temperature",
    "gas viscosity",
    "oil viscosity",
    "gas-oil surface tension",
    "gas mass fraction",
    "gas specific heat",
    "oil specific heat",
    "gas specific enthalpy",
    "oil specific enthalpy",
    "gas thermal conductivity",
    "oil thermal conductivity",
]

if has_water:
    data += [
        water_constant_density_model(p, t).flatten(),
        water_density_derivative_respect_pressure(p, t).flatten(),
        water_density_derivative_respect_temperature(p, t).flatten(),
        water_viscosity_model(p, t).flatten(),
        water_mass_fraction_model(p, t).flatten(),
        water_specific_heat_model(p, t).flatten(),
        water_specific_enthalpy_model(p, t).flatten(),
        water_thermal_conductivity_model(p, t).flatten(),
        gas_water_surface_tension_model(p, t).flatten(),
        oil_water_surface_tension_model(p, t).flatten(),
    ]

    names += [
        "water density",
        "water density derivative pressure",
        "water density derivative temperature",
        "water viscosity",
        "water mass fraction",
        "water specific heat",
        "water specific enthalpy",
        "water thermal conductivity",
        "gas_water_surface_tension",
        "oil_water_surface_tension",
    ]

data.append(temperature(p, h).flatten())
names.append("temperature")
```

src/alfasim_sdk/_internal/alfacase/case_description.py
arthursoprana: Please see if you can avoid using the `.value` and use `attrib_enum` as @prusse-martin suggested. If you run into type-check problems, call any of us for help.

It is better to avoid temporary workarounds in `alfasim-sdk` since this is an open source project and any change here may affect external users of the sdk.

Give it a try again, it seems like a simple change. These type-checking errors may be a simple detail that we're missing, a simple call may solve it ;)

arthursoprana (changes requested):
Nice work! Please take a look at my comments ;)

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: As we discussed in the meeting today we will divide the implementations of the PT and PH table. This way there will be a `TableDescription` for each table and the `table_type` parameter will not be required.

This refactoring will affect different regions of the code and is outside this PR scope. So I will finish it and latter I will a do this refactoring (issue ASIM-4887).

tests/alfacase/test_case_description.py
Gualcar: It is in Kelvin.

arthursoprana (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
prusse-martin: when using `attrib_enum` the schema used to validate the input file will check the valid values, when using a string no special type check is done.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2491,6 +2491,223 @@ def oil_water_surface_tension_model(p, t):
             number_of_phases=3 if has_water else 2,
         )
 
+    @staticmethod
+    def create_constant_ph_table(
```
prusse-martin, line 2495: How is this to be used?

Gualcar: It should be used to create a constant PH table. The funcion `build_constant_pvt_table` in `case_description.py´ will create a constant table depending on the `energy_model_primary_variable` type.

After this PR another one will be created refactoring the PH model. This will probably change this part of the code again.

We discussed that should be better to create an entire `TableDescription` to describre only the `PH tables`.

PR #266, opened 2022-09-21
fb-ASIM-4887 -> Refactor PH table model

Gualcar:
The major idea os this PR is to refactor the PH tables model to keep it independent from the other models. A new TableDescription have been created exclusivly to the PH model.

With this approach the `TableType` enum is no more required, since the table check is done intirely based on the PVT model.

Also, now the `build_constant_pvt_table` build a `PvtModel` description that will fill the `ph` or `pt` table parameter with the right `TableDescription`.

### DIFF
--- a/src/alfasim_sdk/__init__.py
+++ b/src/alfasim_sdk/__init__.py
@@ -134,6 +134,9 @@ from alfasim_sdk._internal.alfacase.case_description import (
     PvtModelTableParametersDescription,
 )
 from alfasim_sdk._internal.alfacase.case_description import (
+    PhPvtModelTableParametersDescription,
+)
+from alfasim_sdk._internal.alfacase.case_description import (
     ReferencedPressureContainerDescription,
 )
 from alfasim_sdk._internal.alfacase.case_description import (
@@ -265,7 +268,6 @@ from alfasim_sdk._internal.constants import PipeThermalModelType
 from alfasim_sdk._internal.constants import PipeThermalPositionInput
 from alfasim_sdk._internal.constants import PumpType
 from alfasim_sdk._internal.constants import PVTCompositionalViscosityModel
-from alfasim_sdk._internal.constants import PVTTableType
 from alfasim_sdk._internal.constants import SeparatorGeometryType
 from alfasim_sdk._internal.constants import SimulationModeType
 from alfasim_sdk._internal.constants import SimulationRegimeType
@@ -440,7 +442,6 @@ __all__ = [
     "OutputAttachmentLocation",
     "OverallPipeTrendDescription",
     "PVTCompositionalViscosityModel",
-    "PVTTableType",
     "PackerDescription",
     "PhysicsDescription",
     "PigEquipmentDescription",
@@ -462,6 +463,7 @@ __all__ = [
     "PvtModelCompositionalDescription",
     "PvtModelCorrelationDescription",
     "PvtModelTableParametersDescription",
+    "PhPvtModelTableParametersDescription",
     "PvtModelsDescription",
     "Quantity",
     "Reference",
--- a/src/alfasim_sdk/__init__.py
+++ b/src/alfasim_sdk/__init__.py
@@ -131,10 +131,10 @@ from alfasim_sdk._internal.alfacase.case_description import (
 )
 from alfasim_sdk._internal.alfacase.case_description import PvtModelsDescription
 from alfasim_sdk._internal.alfacase.case_description import (
-    PvtModelTableParametersDescription,
+    PvtModelPtTableParametersDescription,
 )
 from alfasim_sdk._internal.alfacase.case_description import (
-    PhPvtModelTableParametersDescription,
+    PvtModelPhTableParametersDescription,
 )
 from alfasim_sdk._internal.alfacase.case_description import (
     ReferencedPressureContainerDescription,
@@ -462,8 +462,8 @@ __all__ = [
     "PvtModelCombinedDescription",
     "PvtModelCompositionalDescription",
     "PvtModelCorrelationDescription",
-    "PvtModelTableParametersDescription",
-    "PhPvtModelTableParametersDescription",
+    "PvtModelPtTableParametersDescription",
+    "PvtModelPhTableParametersDescription",
     "PvtModelsDescription",
     "Quantity",
     "Reference",
--- a/src/alfasim_sdk/_internal/alfacase/alfacase.py
+++ b/src/alfasim_sdk/_internal/alfacase/alfacase.py
@@ -26,14 +26,15 @@ def _generate_alfatable_file_for_pvt_models_description(
     """
     from alfasim_sdk import generate_alfatable_file
 
-    for pvt_name, pvt_table_description in pvt_models.table_parameters.items():
+    #TODO: ASIM-4887: generate alfatable files for ph table
+    for pvt_name, pvt_table_description in pvt_models.pt_table_parameters.items():
         alfatable_file = generate_alfatable_file(
             alfacase_file=alfacase_file,
             alfatable_filename=pvt_name,
             description=pvt_table_description,
         )
         pvt_models.tables[pvt_name] = alfatable_file.name
-    pvt_models.table_parameters.clear()
+    pvt_models.pt_table_parameters.clear()
 
 
 def convert_description_to_alfacase(
--- a/src/alfasim_sdk/_internal/alfacase/alfacase.py
+++ b/src/alfasim_sdk/_internal/alfacase/alfacase.py
@@ -26,7 +26,7 @@ def _generate_alfatable_file_for_pvt_models_description(
     """
     from alfasim_sdk import generate_alfatable_file
 
-    #TODO: ASIM-4887: generate alfatable files for ph table
+    #TODO: ASIM-4980: generate alfatable files for ph table
     for pvt_name, pvt_table_description in pvt_models.pt_table_parameters.items():
         alfatable_file = generate_alfatable_file(
             alfacase_file=alfacase_file,
--- a/src/alfasim_sdk/_internal/alfacase/alfacase.py
+++ b/src/alfasim_sdk/_internal/alfacase/alfacase.py
@@ -26,7 +26,7 @@ def _generate_alfatable_file_for_pvt_models_description(
     """
     from alfasim_sdk import generate_alfatable_file
 
-    #TODO: ASIM-4980: generate alfatable files for ph table
+    # TODO: ASIM-4980: generate alfatable files for ph table
     for pvt_name, pvt_table_description in pvt_models.pt_table_parameters.items():
         alfatable_file = generate_alfatable_file(
             alfacase_file=alfacase_file,
--- a/src/alfasim_sdk/_internal/alfacase/case_description.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description.py
@@ -2243,10 +2243,6 @@ class PvtModelTableParametersDescription:
     number_of_phases: int = attr.ib(default=2, validator=instance_of(int))
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
-    table_type: Optional[str] = attr.ib(
-        default=constants.PVTTableType.PT.value, validator=optional(instance_of(str))
-    )
-
     def __attrs_post_init__(self):
         """
         Fix standard properties that have not been set in .TAB files.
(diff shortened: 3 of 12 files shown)


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2492,7 +2488,136 @@ def oil_water_surface_tension_model(p, t):
...
+        validator=numpy_array_validator(dimension=1), repr=collapse_array_repr
+    )
+    table_variables: List[Numpy1DArray] = attr.ib(
+        validator=numpy_array_validator(dimension=1, is_list=True),
+        repr=collapse_array_repr,
+    )
+    variable_names: List[str] = attr.ib(validator=list_of_strings)
+
+    pressure_std = attrib_scalar(default=Scalar(1, "bar"), is_optional=True)
+    temperature_std = attrib_scalar(default=Scalar(15, "degC"), is_optional=True)
```
ggrbill, line 2564: Curiosity:
Since the PvtTable is type `PH`, should it have `enthalpy_std`?

(I am not familiar with PH tables, I am sorry for the dumb question)

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2690,9 +2815,9 @@ def oil_water_surface_tension_model(p, t):
         data.append(temperature(p, h).flatten())
         names.append("temperature")
 
-        return PvtModelTableParametersDescription(
+        return PhPvtModelTableParametersDescription(
             pressure_values=pressure_values,
-            temperature_values=enthalpy_values,
+            enthalpy_values=enthalpy_values,
```
ggrbill, line 2820: Nice! 👍🏼

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2792,6 +2916,15 @@ class PvtModelsDescription:
         where the original PVT file cannot be guaranteed to exist therefore the only reproducible way to recreate
         the PVT is trough the PvtModelTableParametersDescription.
 
+    :ivar ph_table_parameters:
+        *INTERNAL USE ONLY*
+
+        This attribute is populated when exporting a Study to a CaseDescription, and it holds a model representation
+        of a PVT originated from a (.tab / .alfatable) file.
```
ggrbill, line 2923: Don't we have a different pvt table extension? 

Even if we do not have. Perhaps add information about what kind of table this attribute holds. Something like:
```
This attribute is populated when exporting a Study to a CaseDescription, and it holds a model representation
of a PVT originating from a (.tab / .alfatable) file. Also, it holds values from a Pressure-Enthalpy PVT table,
in which temperature is one of the output/state variables.
```

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2829,7 +2962,8 @@ class PvtModelsDescription:
     correlations = attrib_dict_of(PvtModelCorrelationDescription)
     compositional = attrib_dict_of(PvtModelCompositionalDescription)
     combined = attrib_dict_of(PvtModelCombinedDescription)
-    table_parameters = attrib_dict_of(PvtModelTableParametersDescription)
+    pt_table_parameters = attrib_dict_of(PvtModelTableParametersDescription)
+    ph_table_parameters = attrib_dict_of(PhPvtModelTableParametersDescription)
```
ggrbill, line 2966: 👍🏼

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -3173,7 +3308,14 @@ def _get_all_fluids(self) -> List[str]:
...
 
         # Update this tuple when a new PVT Model is added.
-        known_pvt_models = ("compositional", "combined", "correlations", "tables")
+        known_pvt_models = (
+            "compositional",
+            "combined",
+            "correlations",
+            "pt_table_parameters",
+            "ph_table_parameters",
+            "tables",
```
ggrbill, line 3317: Shouldn't `tables` be replaced by two types, like: `pt_tables` and `ph_tables`?
Why did you add the `xx_table_parameters` instead of replacing the original `tables`?

src/alfasim_sdk/_internal/alfacase/schema.py
```diff
@@ -560,6 +580,26 @@
         Optional("surface_tension"): Enum(['Baker Swerdloff']),
     }
 )
+pvt_model_table_parameters_description_schema = Map(
```
ggrbill, line 583: is there a reason to not add `pt_` prefix?
```suggestion
pt_pvt_model_table_parameters_description_schema = Map(
```

ggrbill (approved):
Good Job!
Please take a look at my comments

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: The `enthalpy_std` is calculated inside the `C++` `PhTable` class, but it's calculated from the standard temperature that comes from the table.

This calculation is one of that iterative calculations that we calculate the enthalpy given the temperature.

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: The `tables` attribute is a placeholder (dictionary) to reveice the model name and the `Path` of the table. That way we could use the `tables` attribute to specify a `pt` or `ph` description, for example.

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: For now we have the same suffix `.tab` for `PT` or `PH` tables.

Nice, I will complement the information.

src/alfasim_sdk/_internal/alfacase/schema.py
Gualcar: There is not a reason, since the old name definition was `pvt_model...` to indicate a `pt_pvt_model` I left it unchanged.

ggrbill (approved):
Good work. 

Please, take a look at my comments.

src/alfasim_sdk/_internal/alfacase/schema.py
ggrbill: > For example, the PvtModelTableParametersDescription could be PtPvtModelTableParametersDescription and so on.

I think it is great. It would let things more clear.

ggrbill: > There is not a reason, since the old name definition was `pvt_model...` to indicate a `pt_pvt_model` I left it unchanged.

If you are referring to the general `PvtModel` I think it should keep the name, since it holds any kind of Pvt Models, Like PtTable, PhTable, Composition, ....

However, if it is related to the Original `PvtTable`, I think it should be changed to `PvtPtTable`.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2243,10 +2243,6 @@ class PvtModelTableParametersDescription:
     number_of_phases: int = attr.ib(default=2, validator=instance_of(int))
     warn_when_outside: bool = attr.ib(default=True, validator=instance_of(bool))
 
-    table_type: Optional[str] = attr.ib(
```
prusse-martin, line 2246: Removing this will cause existing alfacase files containing this key to be invalid.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2829,7 +2962,8 @@ class PvtModelsDescription:
     correlations = attrib_dict_of(PvtModelCorrelationDescription)
     compositional = attrib_dict_of(PvtModelCompositionalDescription)
     combined = attrib_dict_of(PvtModelCombinedDescription)
-    table_parameters = attrib_dict_of(PvtModelTableParametersDescription)
```
prusse-martin, line 2832: Renaming an existing property will cause alfacase files using the old key to fail the schema validation.
As of today we don't have any backward compatibility support for alfacase files.

ggrbill: What can we do about it? Is there a way to avoid this problem?

prusse-martin: By "Looks like this was never present in alfacase files."
I was trying to convey that further investigation I think this change will have no actual impact becauseit never was present in alfacase.

prusse-martin: Also `IGNORED_PROPERTIES` in `src/alfasim_sdk/_internal/alfacase/generate_schema.py:274` must be updated.
Remove `table_parameters` and add the now 2 equivalent properties.

Gualcar: I spent a long time yesterday figuring out why the `schema.py` have been updating when I call `inv cog`, and there is the answer. Thank you @prusse-martin!

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: Yes, I have added this attribute before to identify the PH tables, but now it's no more required, after the refactoring.

src/alfasim_sdk/_internal/alfacase/case_description.py
ggrbill: > Also `IGNORED_PROPERTIES` in `src/alfasim_sdk/_internal/alfacase/generate_schema.py:274` must be updated. Remove `table_parameters` and add the now 2 equivalent properties.

Thanks for this information. I didn't remember this detail.

ggrbill: > By "Looks like this was never present in alfacase files." I was trying to convey that further investigation I think this change will have no actual impact becauseit never was present in alfacase.

You are right. Since this property is ignored, it was never present in the `alfacase`.

tests/alfacase/test_alfatable.py
```diff
@@ -61,11 +61,11 @@ def test_alfatable_has_flow_style_for_numpy_array(tmp_path):
           unit: '-'
         number_of_phases: 2
         warn_when_outside: True
-        table_type: pvt_model_table
     """
     )
 
 
+# TODO: ASIM-4887: test the export of a PH table alfacase file
```
Gualcar, line 68: Remember TODOs in the code

src/alfasim_sdk/__init__.py
```diff
@@ -462,6 +463,7 @@ def get_alfasim_sdk_api_path():
     "PvtModelCompositionalDescription",
     "PvtModelCorrelationDescription",
     "PvtModelTableParametersDescription",
+    "PhPvtModelTableParametersDescription",
```
arthursoprana, line 466: I thought that this would be automatically formatted to be added in alphabetical order. It seems that we don't do that. Therefore, could you move this to before `"PhysicsDescription"`, these in theory should be in alphabetical order.

src/alfasim_sdk/_internal/alfacase/alfacase.py
```diff
@@ -26,14 +26,15 @@ def _generate_alfatable_file_for_pvt_models_description(
     """
     from alfasim_sdk import generate_alfatable_file
 
-    for pvt_name, pvt_table_description in pvt_models.table_parameters.items():
+    # TODO: ASIM-4887: generate alfatable files for ph table
```
arthursoprana, line 29: Note that this `TODO` points to the same issue your used for this branch. Do you intend to close the issue right after this merge? If so, this TODO task id should be updated to another issue. If no, nevermind 😄

src/alfasim_sdk/_internal/alfacase/case_description.py
arthursoprana: This was present only in dev versions, it should not be a problem

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2492,7 +2488,136 @@ def oil_water_surface_tension_model(p, t):
...
+        """
+        return PvtModelTableParametersDescription(
+            pressure_values=np.array([0.0]),
+            temperature_values=np.array([0.0]),
+            table_variables=[],
+            variable_names=[],
+            number_of_phases=-1,
+        )
+
+    def __eq__(self, other):
```
arthursoprana, line 2503: I'd suggest to double-check if we use `__eq__` anywhere. If not, we could remove it since this seems to be copied from the original constant pvt table, right?

Simply remove it and run tests (in alfasim and alfasim-sdk), if everything passes, you're all set ;)

tests/alfacase/test_case_description.py
```diff
@@ -788,14 +810,20 @@ def test_pvt_model_table_parameters_description_equal():
     assert table_params_3 != table_params_2
```
arthursoprana, line 810: hmm, that `__eq__` implementation was only there for these test I assume....

arthursoprana (approved):
Nice job! There are some comments I've left, take a look ;)

src/alfasim_sdk/_internal/alfacase/case_description.py
prusse-martin: For that `__eq__` issue take a look at https://www.attrs.org/en/stable/comparison.html#customization maybe we can use the same approach described there.

tests/alfacase/test_case_description.py
Gualcar: I did not find any usage of the equality operator outside this test.

src/alfasim_sdk/_internal/alfacase/case_description.py
Gualcar: Seems to fit in that case -> "This is especially useful when you have fields with objects that have atypical comparison properties. Common examples of such objects are NumPy arrays".

Gualcar: I would try to define these numpy array objects like:
```
pressure_values: Numpy1DArray = attr.ib(
        . . . , eq=attr.cmp_using(eq=numpy.array_equal) 
    )
```
As described in the `attrs` docs shared by @prusse-martin.

prusse-martin: Did not the documented way work?

Gualcar: I had understood that we would change this part if it was not used elsewhere. As the tests are being used [here](https://github.com/ESSS/alfasim-sdk/pull/266/files#diff-43acda733867ce6c06ff05a58287d81d14ec508789ddd167a8402c2e1a954876R778) is it still important to refactor it?

Gualcar: But I think this method will be still needed, since there are an attribute that is a list of numpy arrays.

And we will need the loop over that list:
```
for array1, array2 in zip(self.table_variables, other.table_variables):
    if not np.array_equal(array1, array2):
        return False
```

prusse-martin: `numpy.array_equal` works on lists of ndarrays:
```python
>>> a = [numpy.array([1,2]), numpy.array([3,4])]
>>> b = [numpy.array([1,2]), numpy.array([3,4])]
>>> numpy.array_equal(a, b)
True
>>> b = [numpy.array([1,2]), numpy.array([3,5])]
>>> numpy.array_equal(a, b)
False
```

Gualcar: Great! I will evaluate this change using the approach cited above.

PR #274, opened 2022-11-08
Fb asim 5006 esp viscosity modifications

rafacerq:


ggrbill (approved):

arthursoprana (approved):

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -505,6 +505,7 @@
     ),
     esp_number_of_stages=2,
     esp_reference_density=Scalar(1000.0, "kg/m3"),
+    esp_viscosity_model=constants.PumpViscosityModel.NoModel,
```
prusse-martin, line 508: I think that using the default value here is not good, because if this attribute is ignored in IO the test will pass if we are using the default.

prusse-martin: Until all descriptions are using the auto loaders it is better to stay safe =)
I did check, for `EquipmentDescription` we are using the auto loaders infra so we could look the other way when merging this.

rafacerq: As suggested, the description was removed.

prusse-martin: ehhehe
The problem was not setting it.
The problem was setting it using the same as the default value. Using the other value of `PumpViscosityModel` would be more appropriate.
As I pointed out this description is covered by the autoloaders infra and you don't need to hold back the PR merging for this now.

prusse-martin (approved):
Feel free to resolve our conversation in filled case descriptions =)

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
rafacerq: Sorry for my mistake. Now we have the non-default value: 
`esp_viscosity_model=constants.PumpViscosityModel.AnsiHi2010`

prusse-martin:
The issue for the linux py36 failure: https://github.com/actions/setup-python/issues/544

PR #275, opened 2022-11-08
ASIM-4926-Change esp workflow

BeneBr:
In ASIM-4926 it was necessary to add the EspParameters import in the module's __init__ and configure this enum with auto()

### DIFF
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -363,14 +363,17 @@ class PumpType(Enum):
     TableInterpolation = "table_interpolation"
     ElectricSubmersiblePump = "electric_submersible_pump"
 
+
 class EspParameters(Enum):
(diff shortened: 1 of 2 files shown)


src/alfasim_sdk/_internal/constants.py
```diff
@@ -358,6 +359,11 @@ class MultiInputType(Enum):
     Curve = "curve"
 
 
+class EspParameters(Enum):
+    UserDefined = auto()
+    Catalog = auto()
```
nicoddemus, line 364: As @prusse-martin has commented in Zulip, we should use explicit constants here

```suggestion
class EspParameters(Enum):
    UserDefined = "user_defined"
    Catalog = "catalog"
```

nicoddemus (approved):

prusse-martin (approved):

PR #281, opened 2022-12-01
Force test job for py36/linux to use ubutu 20.04

prusse-martin:
Python 3.6 will no longer be supported on newer ubuntu version since it has reached end-of-live:
https://github.com/actions/setup-python/issues/544#issuecomment-1332535877


dettmerramon (approved):

nicoddemus:
:grimacing: 

Clock is ticking.

PR #284, opened 2022-12-22
Add helper scripts to use `aflasim-sdk` to build plugins using conda env

prusse-martin:


BeneBr (approved):

dettmerramon (approved):

ggrbill:
I would like to say that a description for this PR would clarify why you had to add this feature. 
Perhaps, add a description to register in this PR the reason why you are doing this.

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -120,17 +120,19 @@ def run(self, *, number_of_threads: int = -1) -> Results:
...
-        self._run_simulator(
-            self.get_runner()
-            + [
-                "--run-simulation",
-                f"--alfacase-file={str(self.alfacase_file)}",
-                f"--number-of-threads={number_of_threads}",
-            ],
-            cwd=project_folder,
-        )
+        try:
```
ggrbill, line 123: Doubt:
I know that a generic error would be raised. But what is the name of this error? Which one would be raised?

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -120,17 +120,19 @@ def run(self, *, number_of_threads: int = -1) -> Results:
...
+            self._run_simulator(
```
ggrbill, line 124: You changed this function to return an error code, shouldn't this error code be checked here?

ggrbill (approved):
Please, take a look at my questions.

src/alfasim_sdk/testing/fixtures.py
prusse-martin: `subprocess.CalledProcessError`

src/alfasim_sdk/testing/fixtures.py
prusse-martin: I will not return the error code, just let subprocess do the checking.

tests/results/test_result_reader.py
```diff
@@ -68,7 +73,7 @@ def test_profiles(results: Results) -> None:
 
 
 def test_global_trends(results: Results) -> None:
-    assert results.list_global_trends() == ["timestep"]
+    assert list(map(str, results.list_global_trends())) == ["timestep"]
```
ggrbill, line 76: 👍🏼

PR #285, opened 2023-02-01
Allow alfacase file to omit root models and have empty container models

prusse-martin:


BeneBr (approved):

tests/alfacase/test_plugin_alfacase_to_case.py
```diff
@@ -91,6 +90,54 @@ def test_load_plugin_multiple_containers(datadir, abx_plugin):
     ]
 
 
+def test_load_plugin_missing_top_level_models(datadir: Path, abx_plugin: None) -> None:
```
nicoddemus, line 93: Please add docstrings to these tests explaining what is being tested (TBH I'm not following)

prusse-martin: Really?
OK
```suggestion
def test_load_plugin_missing_top_level_models(datadir: Path, abx_plugin: None) -> None:
    """
    Teas loading plugins with top level models (in `gui_models`) works properly.
    """
```

nicoddemus: Which "top level models" you mean?

Or is that a typo and you meant "loading plugin AS top level models"?

ggrbill: Very minor.

Perhaps "Load plugin with empty data model containers" would let clearer. What do you think?

ggrbill (approved):

tests/alfacase/test_plugin_alfacase_to_case.py
prusse-martin: - [ ] Add docstring;

prusse-martin:
> Perhaps "Load plugin with empty data model containers" would let clearer. What do you think?

If an empty container is present in the alfacase it didn't have a problem loading that.
The problem was loading when containers (or other top level data) are omitted.
And the test fixture dumps a casedescripton to an alfacase, when doing that we will omit empty data structures.
Dumping CaseDescription to alfacase files have not been affected.

"Missing" is the correct term.

tests/alfacase/test_plugin_alfacase_to_case.py
prusse-martin: To avoid any misunderstanding.

> Which "top level models" you mean?

"models" (`data_model`s and/or `container_model`s) named in `gui_models`, the non top level models are not explicitly named in alfacase files.
`container_model`s can only be top level models but `data_model`s can be top level or children of `container_model`s.
The declaration of  `container_model`s explicitly state the type of the contained `data_model`s.

nicoddemus (approved):

PR #288, opened 2023-02-07
Add new options for the ESP heat modelling

rafacerq:
In ASIM-5067, discrepancies were observed when comparing ESP simulation results against experimental data.
Independent of the `thermal_efficiency_constant` values, the simulations were not able to reproduce the steep temperature increase in low flow rate pump conditions.
Therefore, [an alternative ESP heat model](https://drive.google.com/file/d/1IJYX8mMHWpQyhnDRSxnN9osjZFSlbLIl/view?usp=share_link) was implemented in ALFAsim.
Hence, this PR covers the addition of a new Enum (`PumpThermalEfficiencyModel`) to add this option:

The following options were defined:
   - `PumpThermalEfficiencyModel.Constant`: Previous model, where a user define a constant( `thermal_efficiency_constant`) to model the dissipated power;
   - `PumpThermalEfficiencyModel.PumpEfficiencyCurve`: New model, where the constant is defined from the pump efficiency curves;


src/alfasim_sdk/_internal/constants.py
```diff
@@ -418,6 +418,17 @@ class PumpViscosityModel(Enum):
...
 
+class PumpThermalEfficiencyModel(Enum):
+    """
+    Defines the ESP thermal efficiency model
+        - ``Constant``: An user-defined constant value is used to define to transform the ESP mechanical energy to heat.
+        - ``PumpEfficiencyCurve``: The ESP efficiency curve is used to transform the ESP mechanical energy to heat.
+    """
+
+    Constant = "constant"
+    PumpEfficiencyCurve = "pump_eff_curve"
```
prusse-martin, line 429: Is `pump_eff_curve` the best term? 
Do you think `pump_efficiency_curve` is to large?
What about `adjusted_efficiency_curve`, or `efficiency_curve_based`, or `efficiency_based`? (something in those lines)

Remember that this string will be present in alfacase files.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -762,6 +762,9 @@ class PumpEquipmentDescription:
     esp_parameters = attrib_enum(default=constants.EspParameters.UserDefined)
 
     esp_viscosity_model = attrib_enum(default=constants.PumpViscosityModel.NoModel)
+    thermal_efficiency_model = attrib_enum(
+        default=constants.PumpThermalEfficiencyModel.Constant
+    )
```
prusse-martin, line 767: From
```python
class PumpThermalEfficiencyModel(Enum):
    """
    Defines the ESP thermal efficiency model
```
I believe this should be
```suggestion
    esp_thermal_efficiency_model = attrib_enum(
        default=constants.PumpThermalEfficiencyModel.Constant
    )
```

BTW.
Are non ESP pumps affected by this option?
If they are the enum's docstring should be updated instead.

prusse-martin (changes requested):
Remember to update the changelog.

src/alfasim_sdk/_internal/alfacase/case_description.py
rafacerq: Only the ESP pumps are affected by this option. Therefore, I followed you suggestion and renamed it to `esp_thermal_efficiency_model `.

src/alfasim_sdk/_internal/constants.py
rafacerq: Following your suggestion, I renamed it to `efficiency_curve_based`.

rafacerq:
> Remember to update the changelog.

Sure! The changelog is update accordingly.

prusse-martin (approved):

PR #289, opened 2023-02-08
Allow ALFAsimTestingFixture to run simulations from CaseDescription 

ggrbill:
Since internally already exists a conversion from `alfacase` to `CaseDescription`, in order for the test case to be updated, it is a good approach to be possible to run a test case directly from a `CaseDescription`.

It also improves the simulation running process, which will be able to be performed by our two available text-based test case configurations.

### DIFF
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -234,6 +234,8 @@ class MaterialType(Enum):
 FLUID_GAS = "gas"
 FLUID_OIL = "oil"
 FLUID_WATER = "water"
+FLUID_DROPLET = "droplet"
+FLUID_BUBBLE = "bubble"
 FLUID_PHASE_NAMES = [FLUID_GAS, FLUID_OIL, FLUID_WATER]
 
 
--- a/src/alfasim_sdk/testing/fixtures.py
+++ b/src/alfasim_sdk/testing/fixtures.py
@@ -66,7 +66,10 @@ class AlfasimRunnerFixture:
         self.check_not_run()
         self.check_base_case_is_not_loaded()
 
-        self._case = convert_alfacase_to_description(Path(*parts))
+        if len(parts) == 1 and isinstance(parts[0], CaseDescription):
+            self._case = parts[0]
+        else:
+            self._case = convert_alfacase_to_description(Path(*parts))
 
     def add_plugin_folder(self, *parts: Union[str, Path]) -> None:
         """
--- a/src/alfasim_sdk/__init__.py
+++ b/src/alfasim_sdk/__init__.py
@@ -242,6 +242,8 @@ from alfasim_sdk._internal.constants import FlowDirection
 from alfasim_sdk._internal.constants import FlowPatternModel
 from alfasim_sdk._internal.constants import FLUID_GAS
 from alfasim_sdk._internal.constants import FLUID_OIL
+from alfasim_sdk._internal.constants import FLUID_DROPLET
+from alfasim_sdk._internal.constants import FLUID_BUBBLE
 from alfasim_sdk._internal.constants import FLUID_PHASE_NAMES
 from alfasim_sdk._internal.constants import FLUID_WATER
 from alfasim_sdk._internal.constants import GAS_FIELD
@@ -389,6 +391,8 @@ __all__ = [
     "EvaluationStrategyType",
     "FLUID_GAS",
     "FLUID_OIL",
+    'FLUID_BUBBLE',
+    'FLUID_DROPLET',
     "FLUID_PHASE_NAMES",
     "FLUID_WATER",
     "FileContent",


src/alfasim_sdk/_internal/constants.py
```diff
@@ -234,6 +234,8 @@ class MaterialType(Enum):
 FLUID_GAS = "gas"
 FLUID_OIL = "oil"
 FLUID_WATER = "water"
+FLUID_DROPLET = "droplet"
+FLUID_BUBBLE = "bubble"
```
arthursoprana, line 238: Are you sure this is required? Note that below we have `FLUID_PHASE_NAMES = [FLUID_GAS, FLUID_OIL, FLUID_WATER]`

It seems that `FLUID_DROPLET` and `FLUID_BUBBLE` are old names that should not be exposed in the `alfasim-sdk`. Couldn't you add these hard-coded constants directly in the `alfasim_plugin_hydrates`?

It seems that the general user **should not** have access to `FLUID_DROPLET` and `FLUID_BUBBLE`

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -66,7 +66,10 @@ def load_base(self, *parts: Union[str, Path]) -> None:
         self.check_not_run()
         self.check_base_case_is_not_loaded()
 
-        self._case = convert_alfacase_to_description(Path(*parts))
+        if len(parts) == 1 and isinstance(parts[0], CaseDescription):
```
arthursoprana, line 69: The docstring `Define the base case from an alfacase file.` for this method may need to be updated since it may now receive either a `CaseDescription` or an `alfacase`

tests/fixtures/test_alfasim_runner.py
```diff
@@ -70,3 +70,29 @@ def test_alfasim_runner_add_plugin(
...
+
+def test_alfasim_runner_from_case_description(
+    alfasim_runner: AlfasimRunnerFixture,
+    fake_alfasim: None,
+    empty_alfacase: Path,
+    abx_plugin_source: Path,
+) -> None:
+    from alfasim_sdk import convert_alfacase_to_description
+
+    case_description = convert_alfacase_to_description(Path(empty_alfacase))
```
arthursoprana, line 83: does `convert_alfacase_to_description` returns a list?

src/alfasim_sdk/testing/fixtures.py
arthursoprana: Moreover, this ideally should be

```python
def load_base(self, *parts: Union[str, Path], case_description: Optional[CaseDescription] = None) -> None:
```

and then

```python
if case_description is not None:
    self._case = case_description
else:
    self._case = convert_alfacase_to_description(Path(*parts))
```

The current implementation does not make it explicit that `load_base` can receive a `CaseDescription`.

arthursoprana (changes requested):

src/alfasim_sdk/_internal/constants.py
ggrbill: Since the plugin developer would use directly the strings `"droplet"` or `"bubble"`, I think it is nice to create a tag and make it available to everyone. 
Note that it could be used to change the layers and phases by a plugin in order to create a specific MultiFieldDescription. 

Do you think it could be useful?

src/alfasim_sdk/testing/fixtures.py
ggrbill: Nice suggestion. 

- [ ] Update docstring
- [ ] Add new parameter `case_description`

tests/fixtures/test_alfasim_runner.py
ggrbill: No, it returns a `CaseDescription` object.

src/alfasim_sdk/_internal/constants.py
prusse-martin: "water in oil" is a bubble or droplet? and what about "oil in water"?

And what if a plugin add a new phase? How does the new phase interact (declare how interact?) with other phases to create droplets and bubbles?

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -59,14 +59,21 @@ def check_base_case_is_loaded(self) -> None:
...
+        Define the base case from an alfacase file or a CaseDescription.
         """
         self.check_not_run()
         self.check_base_case_is_not_loaded()
 
-        self._case = convert_alfacase_to_description(Path(*parts))
+        if case_description is not None:
+            self._case = case_description
+        else:
+            self._case = convert_alfacase_to_description(Path(*parts))
```
prusse-martin, line 76: Properly typing this and preventing it to be wrongly used can be tricky.
Maybe we should be more explicit.

```python
    def check_can_load_base(self) -> None:
        self.check_not_run()
        self.check_base_case_is_not_loaded()

    def load_base_from_alfacase(self, path: Union[str, Path]) -> None:
        self.check_can_load_base()
        self._case = convert_alfacase_to_description(Path(path))
    
    def load_base_from_case_description(self, case_description: CaseDescription) -> None:
        self.check_can_load_base()
        self._case = case_description
```

(docstring omitted for the benefit of a short example)

tests/fixtures/test_alfasim_runner.py
```diff
@@ -70,3 +70,29 @@ def test_alfasim_runner_add_plugin(
...
+    alfasim_runner.load_base(case_description=case_description)
+    alfasim_runner.add_plugin_folder(abx_plugin_source)
```
prusse-martin, line 85: I plan on removing `add_plugin_folder` from the fixture, use the `abx_plugin` to have the same effect.

prusse-martin (changes requested):

src/alfasim_sdk/_internal/constants.py
ggrbill: > "water in oil" is a bubble or droplet?

None of them.

> and what about "oil in water"?

In this situation, it is not needed.

> And what if a plugin add a new phase? How does the new phase interact (declare how interact?) with other phases to create droplets and bubbles?

A plugin can update a phase or a layer by adding a new field to it. And if the MFD used by the user has no bubble or droplet, it can be added by the plugin to have the complete MFD needed to support its models.

We make something similar with the water phase in the hydrates plugin. We ensure the user do not choose a MFD with water and the plugin adds a new phase of water as a dispersed one updating the oil layer (different from our own triphase MFD which has free water, or layer water)

src/alfasim_sdk/testing/fixtures.py
ggrbill: I think is a better idea. I will change it.

- [ ] Create explicit `load_case` function fro `alfacase` and `CaseDescription`

tests/fixtures/test_alfasim_runner.py
ggrbill: I will let it as it is because there are more tests using it.
Is that ok?

src/alfasim_sdk/_internal/constants.py
arthursoprana: The main issue here is that we have just down below

```python
FLUID_PHASE_NAMES = [FLUID_GAS, FLUID_OIL, FLUID_WATER]
```


and now if I'm a new guy trying to understand the sdk I see

```python
FLUID_GAS = "gas"
FLUID_OIL = "oil"
FLUID_WATER = "water"
FLUID_DROPLET = "droplet"
FLUID_BUBBLE = "bubble"
```

what are these `FLUID_DROPLET ` and `FLUID_BUBBLE `? phases? fields?

@prusse-martin's questions are relevant. If he is already confused about it, imagine someone completely new. `FLUID_BUBBLE` today is meant to be used by a specific subset of Multi Field Descriptions (e.g., four-phase). I'd rather keep it hard coded in the plugin side for now and add a more consistent solution (that servers for all scenarios, two/three/N-phases) once we've got one.

src/alfasim_sdk/testing/fixtures.py
arthursoprana: did you commit the code? It seems like the original one

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -59,15 +59,26 @@ def check_base_case_is_loaded(self) -> None:
...
         self.check_base_case_is_not_loaded()
 
+    def load_base_from_alfacase(self, *parts: Union[str, Path]) -> None:
+        """
+        Define the base case from an alfacase file.
+        """
+        self.check_can_load_base()
         self._case = convert_alfacase_to_description(Path(*parts))
 
+    def load_base_from_case_description(
```
arthursoprana, line 73: 👍

src/alfasim_sdk/_internal/constants.py
arthursoprana: I suggest to remove `FLUID_DROPLET` and `FLUID_BUBBLE`

arthursoprana (approved):

src/alfasim_sdk/_internal/constants.py
ggrbill: OK. I understand your concern.
I wasn't so worried about it. But you have a point.

- [ ] I will move it to the plugin and change it to have the `FIELD_` prefix instead of `FLUID_`.

tests/fixtures/test_alfasim_runner.py
prusse-martin: OK.

src/alfasim_sdk/_internal/constants.py
prusse-martin: I was thinking that `*_DROPLET` was an alias to "water in air" and `*_BUBBLE` an alias to "air in water".

prusse-martin (changes requested):
Remember to update the change log given you are adding that bubble droplet stuff to the interface.

ggrbill:
> Remember to update the change log given you are adding that bubble droplet stuff to the interface.

Not anymore. I completely removed those stuffs.

prusse-martin (approved):

PR #293, opened 2023-02-14
[PYTHON 3.10] Refactored pyyaml code in order to avoid deprecated api

arthursoprana:
- Simple refactoring to avoid deprecated api (which will give a warning msg in python 3.10)

EDEN-2592

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/alfatable.py
+++ b/src/alfasim_sdk/_internal/alfacase/alfatable.py
@@ -33,11 +33,12 @@ def load_pvt_model_table_parameters_description_from_alfatable(
     """
     Load the content from the alfatable in the given file_path. The validation is turned off due to performance issues.
     """
-    from ruamel import yaml as ruamelyaml
+    from ruamel.yaml.main import YAML
     from barril.units import Scalar
     import numpy as np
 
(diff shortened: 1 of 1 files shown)


src/alfasim_sdk/_internal/alfacase/alfatable.py
```diff
@@ -33,11 +33,12 @@ def load_pvt_model_table_parameters_description_from_alfatable(
     """
     Load the content from the alfatable in the given file_path. The validation is turned off due to performance issues.
     """
-    from ruamel import yaml as ruamelyaml
+    from ruamel.yaml.main import YAML
     from barril.units import Scalar
     import numpy as np
 
-    content = ruamelyaml.safe_load(Path(file_path).read_text(encoding="UTF-8"))
```
prusse-martin, line 40: Don't you need to set an encoding?
Did you try to open a alfacase with utf8 encoding?

prusse-martin (changes requested):

src/alfasim_sdk/_internal/alfacase/alfatable.py
arthursoprana: Well it seems that we're lacking tests for "open a alfacase with utf8 encoding", because all tests passed with this change.

Anyway since this is minor I'll tackle it in another moment.

PR #294, opened 2023-02-23
ASIM-5108-Fix reference density behavior in ESP

BeneBr:
Was update the alfacase to have the enabled_density_correction variable.

[ASIM-5108](https://eden.esss.co/jira/browse/ASIM-5108)

### DIFF
--- a/docs/source/alfacase_definitions/PumpEquipmentDescription.txt
+++ b/docs/source/alfacase_definitions/PumpEquipmentDescription.txt
@@ -8,6 +8,7 @@
             position: \ :class:`Scalar <barril.units.Scalar>`
             flow_direction: \ :class:`FlowDirection <alfasim_sdk._internal.constants.FlowDirection>`\  = FlowDirection.Forward
             thermal_efficiency: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(100.0, '%', 'dimensionless')
+            thermal_efficiency_model: \ :class:`PumpThermalEfficiencyModel <alfasim_sdk._internal.constants.PumpThermalEfficiencyModel>`\  = PumpThermalEfficiencyModel.Constant
             type: \ :class:`PumpType <alfasim_sdk._internal.constants.PumpType>`\  = PumpType.ConstantPressure
             pressure_boost: \ :class:`Scalar <barril.units.Scalar>`\  = Scalar(100000.0, 'Pa', 'pressure')
             table: \ :class:`TablePumpDescription <TablePumpDescription>`\  = TablePumpDescription()
@@ -22,7 +23,7 @@
             user_defined_esp_table: \ :class:`TablePumpDescription <TablePumpDescription>`\  = TablePumpDescription()
             esp_parameters: \ :class:`EspParameters <alfasim_sdk._internal.constants.EspParameters>`\  = EspParameters.UserDefined
             esp_viscosity_model: \ :class:`PumpViscosityModel <alfasim_sdk._internal.constants.PumpViscosityModel>`\  = PumpViscosityModel.NoModel
-            esp_thermal_efficiency_model: \ :class:`PumpThermalEfficiencyModel <alfasim_sdk._internal.constants.PumpThermalEfficiencyModel>`\  = PumpThermalEfficiencyModel.Constant
+            density_correction_enabled: bool = False
 
 .. tab:: Schema
 
@@ -35,6 +36,7 @@
             thermal_efficiency:  # optional
                 value: number
                 unit: string
+            thermal_efficiency_model: \ :class:`PumpThermalEfficiencyModel <alfasim_sdk._internal.constants.PumpThermalEfficiencyModel>`\  # optional
             type: \ :class:`PumpType <alfasim_sdk._internal.constants.PumpType>`\  # optional
             pressure_boost:  # optional
                 value: number
@@ -61,4 +63,4 @@
             user_defined_esp_table: \ :class:`table_pump_description_schema <TablePumpDescription>`\  # optional
             esp_parameters: \ :class:`EspParameters <alfasim_sdk._internal.constants.EspParameters>`\  # optional
             esp_viscosity_model: \ :class:`PumpViscosityModel <alfasim_sdk._internal.constants.PumpViscosityModel>`\  # optional
-            esp_thermal_efficiency_model: \ :class:`PumpThermalEfficiencyModel <alfasim_sdk._internal.constants.PumpThermalEfficiencyModel>`\  # optional
+            density_correction_enabled: boolean  # optional
--- a/src/alfasim_sdk/_internal/alfacase/case_description.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description.py
@@ -726,6 +726,7 @@ class PumpEquipmentDescription:
     position = attrib_scalar(category="length")
     flow_direction = attrib_enum(default=constants.FlowDirection.Forward)
     thermal_efficiency = attrib_scalar(default=Scalar(100.0, "%"))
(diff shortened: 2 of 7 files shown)


cauebs (approved):

prusse-martin (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -765,6 +765,9 @@ class PumpEquipmentDescription:
     esp_thermal_efficiency_model = attrib_enum(
         default=constants.PumpThermalEfficiencyModel.Constant
     )
+    enabled_density_correction: bool = attr.ib(
+        default=False, validator=instance_of(bool)
+    )
```
prusse-martin, line 770: Why `enabled_density_correction`?
Probably `enable_density_correction` is better.

BeneBr: Ok. i change this.

prusse-martin (approved):

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -507,6 +507,7 @@
     esp_reference_density=Scalar(1000.0, "kg/m3"),
     user_defined_esp_table=TABLE_PUMP_DESCRIPTION,
     esp_parameters=constants.EspParameters.Catalog,
+    enable_density_correction=False,
```
prusse-martin, line 510: You probably want to chance the `esp_thermal_efficiency_model`. Even disabled it will be dumped, and it is important that to check this.

prusse-martin (changes requested):

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
BeneBr: Can you be more especific about that?

prusse-martin: Set the value of `esp_thermal_efficiency_model` when creating this case description to a non default value.

prusse-martin: does this comes from master?
```
class PumpThermalEfficiencyModel(Enum):
    """
    Defines the ESP thermal efficiency model
```

@Gualcar 
Is this eps only related? Shouldn't ne prefix of with `eps_` to make that clear?

Gualcar: I think @rafacerq will know it properly. But, this `PumpThermalEfficiencyModel` is an attribute of the class `Pump` (the parent class for all pumps), so I think it should be maintained as general term.

prusse-martin: If that is the case the enum doc string must be updated.

Gualcar: Yeah, it could be replaced by `Defines the pump thermal efficiency model`.

src/alfasim_sdk/_internal/constants.py
```diff
@@ -418,6 +418,17 @@ class PumpViscosityModel(Enum):
...
 
+class PumpThermalEfficiencyModel(Enum):
+    """
+    Defines the pump thermal efficiency model
+        - ``Constant``: An user-defined constant value is used to define to transform the ESP mechanical energy to heat.
+        - ``PumpEfficiencyCurve``: The ESP efficiency curve is used to transform the ESP mechanical energy to heat.
+    """
+
+    Constant = "constant"
+    PumpEfficiencyCurve = "pump_eff_curve"
```
prusse-martin, line 429: Use the long name. it is clearer, and that is the value that ends in alfacase files.
```suggestion
    PumpEfficiencyCurve = "pump_efficiency_curve"
```

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
prusse-martin, line 510: Now we are missing `thermal_efficiency_model`.

prusse-martin (changes requested):

src/alfasim_sdk/_internal/constants.py
rafacerq: Is this change necessary? We already have ``PumpEfficiencyCurve = "efficiency_curve_based"`` at the master.

prusse-martin: You are right.
Probably using the one from master is better.

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -762,8 +765,8 @@ class PumpEquipmentDescription:
     esp_parameters = attrib_enum(default=constants.EspParameters.UserDefined)
 
     esp_viscosity_model = attrib_enum(default=constants.PumpViscosityModel.NoModel)
-    esp_thermal_efficiency_model = attrib_enum(
-        default=constants.PumpThermalEfficiencyModel.Constant
+    enable_density_correction: bool = attr.ib(
```
nicoddemus, line 768: "enable" here is a verb, this reads better:

```suggestion
    density_correction_enabled: bool = attr.ib(
```

src/alfasim_sdk/_internal/constants.py
```diff
@@ -420,7 +420,18 @@ class PumpViscosityModel(Enum):
 
 class PumpThermalEfficiencyModel(Enum):
```
nicoddemus, line 421: Why is this class defined twice? :thinking:

nicoddemus (changes requested):

docs/source/alfacase_definitions/PumpEquipmentDescription.txt
```diff
@@ -22,7 +23,7 @@
             user_defined_esp_table: \ :class:`TablePumpDescription <TablePumpDescription>`\  = TablePumpDescription()
             esp_parameters: \ :class:`EspParameters <alfasim_sdk._internal.constants.EspParameters>`\  = EspParameters.UserDefined
             esp_viscosity_model: \ :class:`PumpViscosityModel <alfasim_sdk._internal.constants.PumpViscosityModel>`\  = PumpViscosityModel.NoModel
-            esp_thermal_efficiency_model: \ :class:`PumpThermalEfficiencyModel <alfasim_sdk._internal.constants.PumpThermalEfficiencyModel>`\  = PumpThermalEfficiencyModel.Constant
```
rafacerq, line 25: Why it was removed? It is currently being used to enable the thermal efficiency model at master.

rafacerq (changes requested):
Nice work!
Take a look at my comments, it seems that `esp_thermal_efficiency_model` is missing.

docs/source/alfacase_definitions/PumpEquipmentDescription.txt
BeneBr: @rafacerq yet i see this. So this attribute is used for all pump types, so they now will be renamed by ```thermal_efficiency_model```

src/alfasim_sdk/_internal/constants.py
BeneBr: One of these class was removed. It was a mistake.

docs/source/alfacase_definitions/PumpEquipmentDescription.txt
rafacerq: Ok!

rafacerq (approved):

nicoddemus (approved):

PR #296, opened 2023-03-17
Add log to alfasim_run when simulation fails

ThalesCarl:
I've added a exception handler in the `run` method in order to have more detailed info when the simulation fails. Before this change, the execution only returned the `subprocess` exit code.

ASIM-3405


src/alfasim_sdk/testing/fixtures.py
```diff
@@ -162,6 +163,10 @@ def run(self, *, number_of_threads: int = -1) -> Results:
                 ],
                 cwd=project_folder,
             )
+        except CalledProcessError:
+            incorrect_results = Results(self.alfacase_data_folder)
+            if incorrect_results is not None:
+                print(incorrect_results.log.read_text())
```
ggrbill, line 169: Perhaps print the `log_calc` file also.
What do you think, @prusse-martin ?

ggrbill (approved):

ThalesCarl:
I've seen that the coverage degraded with my addition. Hence, I am adding one test to cover the scenario where a CalledProcessError is raised. So far, I've duplicated the code of the [test_alfasim_runner_from_case_description](https://github.com/ESSS/alfasim-sdk/blob/master/tests/fixtures/test_alfasim_runner.py#L75) and I've tried adding a pipe with a invalid pvt_model, as shown below. However, the simulation runs without raising the CalledProcessError. Do you guys know what could I be missing here?

```python
pipe = PipeDescription(
        'pipe',
        'source',
        'target',
        pvt_model='unknown',
        segments=PipeSegmentsDescription(
            start_positions=Array([0.0], 'm'),
            diameters=Array([0.1], 'm'),
            roughnesses=Array([1e-5], 'm'),
        ),
    )
    case_description = attr.evolve(
        case_description,
        pipes=[pipe]
    )
```

src/alfasim_sdk/testing/fixtures.py
prusse-martin: I have thought  that `log_calc` is not created when the simulation fail.

prusse-martin:
>

prusse-martin (commented):
See the comment I put in the PR.

src/alfasim_sdk/testing/fixtures.py
ThalesCarl: It creates the log_calc file also, at least for the cases I have seen. I have added the log_calc.txt content to this printing. But I want your opinion if does it makes sense, because the log_calc.txt output can be large depending on the situation.

tests/fixtures/test_alfasim_runner.py
```diff
@@ -96,3 +115,20 @@ def test_alfasim_runner_from_case_description(
...
+) -> None:
+    from alfasim_sdk import convert_alfacase_to_description
+
+    case_description = convert_alfacase_to_description(Path(empty_alfacase))
+
+    alfasim_runner.load_base_from_case_description(case_description)
+    fake_alfasim_return_code(1)
+    with pytest.raises(RuntimeError) as exp:
+        alfasim_runner.run()
+        assert str(exp.value) != ""
```
prusse-martin, line 134: I think it is better to check for some part of the log here, not just `!= ""`.

```suggestion
    with pytest.raises(RuntimeError, match="run_simulation.*Reading study"):
        alfasim_runner.run()
```

prusse-martin (changes requested):

tests/fixtures/test_alfasim_runner.py
```diff
@@ -96,3 +115,20 @@ def test_alfasim_runner_from_case_description(
...
+    alfasim_runner: AlfasimRunnerFixture,
+    fake_alfasim: None,
+    empty_alfacase: Path,
+    fake_alfasim_return_code,
+) -> None:
+    from alfasim_sdk import convert_alfacase_to_description
+
+    case_description = convert_alfacase_to_description(Path(empty_alfacase))
+
+    alfasim_runner.load_base_from_case_description(case_description)
```
prusse-martin, line 130: Is there coverage issues?
Why not just

```suggestion
) -> None:
    alfasim_runner.load_base_from_alfacase(empty_alfacase)
```

ThalesCarl: They are not. I just had copied the previous test and didn't notice I could have used directly the `load_base_from_alfacase` function. Thanks for the suggestion.

src/alfasim_sdk/testing/fixtures.py
ThalesCarl: @ggrbill what do you think about including the log_calc.txt in this exception handling?

ggrbill: I think it is not a problem, since this file is short when the simulation fails. In most cases the simulation will fail right in the beginning, so the file could be printed.

src/alfasim_sdk/testing/fixtures.py
```diff
@@ -162,6 +163,21 @@ def run(self, *, number_of_threads: int = -1) -> Results:
...
+                f"""\
+                ========================================
+                === log_calc.txt
+                {log_calc.read_text() if log_calc.is_file() else "<not found>"}
+                ========================================
+                === log.txt
+                {log.read_text() if log.is_file() else "<not found>"}
+                """
+            )
+            raise RuntimeError(msg) from error
```
prusse-martin, line 180: Don't need to change.
I just update something very similar to this in alfasim.
It will be something like:

```suggestion
            def list_log(log_file):
                try:
                    contents = log_file.read_text()
                except Exception as e:
                    contents = f'<ERROR READING FILE: {str(e)}>'

                name = log_file.name
                header_len = 4 + len(name) + 4
                return f'{"=" * header_len}\n=== {log_file.name} ===\n{"=" * header_len}\n{contents}'
                
            results = Results(self.alfacase_data_folder)
            log_calc = results.log_calc
            log = results.log
            msg = '\n'.join([list_log(log_calc), list_log(log)])
            raise RuntimeError(msg) from error
```

prusse-martin (approved):

PR #297, opened 2023-03-24
Fix method `to_curve`

Gualcar:
This PR fix the method `to_curve` to return a curve if the length of `value` parameter is 0.

Tests from `test_configure_case_from_study.py` were crashing because the `GetAsCurve()` was returning nothing and the method `to_curve` did not know how to handle it.

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
@@ -193,7 +193,10 @@ def to_curve(
         return Curve(image, domain)
     elif isinstance(value, Curve):
         if value.GetLength() == 0:
-            return Curve(image=Array([1.0], value.GetImage().unit), domain=Array([1.0], value.GetDomain().unit))
+            return Curve(
+                image=Array([1.0], value.GetImage().unit),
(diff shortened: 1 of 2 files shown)


src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
```diff
@@ -192,6 +192,11 @@ class Foo:
         )
         return Curve(image, domain)
     elif isinstance(value, Curve):
+        if value.GetLength() == 0:
+            return Curve(
+                image=Array([1.0], value.GetImage().unit),
+                domain=Array([1.0], value.GetDomain().unit),
+            )
```
prusse-martin, line 199: This is not good =/
We should not create data from thin air.

prusse-martin: Huh this is no good, why are we returning a curve with a single point using this arbitrary values? If somewhere else we are getting an error about the curve being empty, we should fix the error there, not just returning an arbitrary curve like this one.

nicoddemus: cc @Gualcar @BeneBr

PR #301, opened 2023-04-10
Disable HDF5 file locks

prusse-martin:


nicoddemus (approved):

src/alfasim_sdk/_internal/alfacase/case_to_alfacase.py
```diff
@@ -38,11 +38,9 @@ def format_list(values: List[Any], *, enable_flow_style: bool = False):
     For more details check:
     https://stackoverflow.com/questions/63364894/how-to-dump-only-lists-with-flow-style-with-pyyaml-or-ruamel-yaml
     """
-    import ruamel
+    import strictyaml
 
-    # TODO: arthur: Review this workaround before merging!!!
-    return values
-    retval = ruamel.yaml.comments.CommentedSeq(values)
+    retval = strictyaml.ruamel.comments.CommentedSeq(values)
```
prusse-martin, line 43: @arthursoprana 
I did not like the workaround.

prusse-martin: Ooops.
I am using a different version =)

prusse-martin: First. I did undo this and add a xfail with a TODO in the affected test.
The using the conda version of the package works this way but the one obtained using pip do not.
The problem with using `CommentedSeq` is possibly due stricyaml now vendoring ruamel or using their own implementation (and later checking `isinstance`).

The original code will cause the dumpoed yaml do be formatted as:
```
a: [1, 2, 3]
```
with this workaround introduced by @arthursoprana:
```
a:
  - 1
  - 2
  - 3
```
Note that I use 3 items in the list as an example, I don't know the expected item count in real cases.

nicoddemus: If this fixes it for now, I'm all up for using the hack (we just need to leave a comment explaining the problem, and possible future solutions, like waiting for a new release).

prusse-martin:
```
    # The strictyaml version used by alfasim has ruamel vendored
    # making it unable to work with the "non-vendored" ruamel version.
    # The strictyaml version obtained with pip to run test on github
    # uses a non vendored version of ruamel.
    # By now we use by defaul here the vendored ruamel with a fallback
    # to the original ruamel.
    # When updating alfasim's strictyaml we probably should review this hack.
```

nicoddemus: Great!

PR #302, opened 2023-04-10
[PYTHON 3.10] Update alfasim to python 310

arthursoprana:


src/alfasim_sdk/_internal/alfacase/case_to_alfacase.py
```diff
@@ -40,6 +40,8 @@ def format_list(values: List[Any], *, enable_flow_style: bool = False):
     """
     import ruamel
 
+    # TODO: arthur: Review this workaround before merging!!!
+    return values
     retval = ruamel.yaml.comments.CommentedSeq(values)
```
prusse-martin, line 45: I think this will work.
```suggestion
    import strictyaml

    if hasattr(strictyaml, 'ruamel'):
        ruamel = strictyaml.ruamel
    else:
        import ruamel

    retval = ruamel.yaml.comments.CommentedSeq(values)
```

Note that the original code will work with ruamel/stricyaml from pypi, it fail on alfasim because the package/version of strictyaml is vendoring ruamel.
This code uses the vendored ruamel version if it is being vendored and falls back to the original ruamel.

Probably a better solution is updating the package we are using on alfasim.

prusse-martin: I will add this to
https://github.com/ESSS/alfasim-sdk/pull/301

src/alfasim_sdk/_internal/alfacase/case_to_alfacase.py
```diff
@@ -44,15 +44,16 @@ def format_list(values: List[Any], *, enable_flow_style: bool = False):
     # making it unable to work with the "non-vendored" ruamel version.
     # The strictyaml version obtained with pip to run test on github
     # uses a non vendored version of ruamel.
-    # By now we use by defaul here the vendored ruamel with a fallback
+    # By now we use by default here the vendored ruamel with a fallback
     # to the original ruamel.
     # When updating alfasim's strictyaml we probably should review this hack.
     if hasattr(strictyaml, "ruamel"):
-        ruamel = strictyaml.ruamel
+        comments = strictyaml.ruamel.comments
```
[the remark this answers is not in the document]
nicoddemus: Can you please do a final review?

src/alfasim_sdk/_internal/alfacase/alfatable.py
```diff
@@ -33,11 +33,12 @@ def load_pvt_model_table_parameters_description_from_alfatable(
...
     Load the content from the alfatable in the given file_path. The validation is turned off due to performance issues.
     """
-    from ruamel import yaml as ruamelyaml
+    from ruamel.yaml.main import YAML
     from barril.units import Scalar
     import numpy as np
 
-    content = ruamelyaml.safe_load(Path(file_path).read_text(encoding="UTF-8"))
+    yaml = YAML(typ="safe", pure=True)
+    content = yaml.load(Path(file_path))
```
prusse-martin, line 41: @arthursoprana 
Without `pure=True` did it fail?
```
pure: if True only use Python modules
```
I think using `False` could be faster.

prusse-martin (approved):

PR #309, opened 2023-05-18
Add Vogel IPR model descriptions

caio-ramos:
ASIM-3662

Add a `CaseDescription` of the Vogel IPR model. 

The ALFAsim PR counterpart is available [here](https://github.com/ESSS/alfasim/pull/271)


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1118,6 +1118,30 @@ class LinearIPRDescription(CommonIPR):
...
+
+    # [[[cog
+    # cog_out_multi_input("well_max_flow_rate", "standard volume per time", 0.0, "sm3/d")
+    # ]]]
+    # fmt: off
+    well_max_flow_rate_input_type = attrib_enum(default=constants.MultiInputType.Constant)
+    well_max_flow_rate = attrib_scalar(
+        default=Scalar('standard volume per time', 0.0, 'sm3/d')
+    )
+    well_max_flow_rate_curve = attrib_curve(
```
ggrbill, line 1140: Does Vogel's model consider a table of Max Flow Rate?
 I am asking because when we change the Max flow rate we are changing Vogel's model behavior. I mean, we are changing the entire IPR curve of the model during the simulation. Does it make sense? Did you find a reference that presented this kind of approach?

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1118,6 +1118,30 @@ class LinearIPRDescription(CommonIPR):
...
+
+    min_pressure_difference = attrib_scalar(default=Scalar(0.0, "Pa"))
+
+    # [[[cog
+    # cog_out_multi_input("well_max_flow_rate", "standard volume per time", 0.0, "sm3/d")
+    # ]]]
+    # fmt: off
+    well_max_flow_rate_input_type = attrib_enum(default=constants.MultiInputType.Constant)
+    well_max_flow_rate = attrib_scalar(
+        default=Scalar('standard volume per time', 0.0, 'sm3/d')
```
ggrbill, line 1138: According to Vogel's model, the Max Flow Rate is used to calculate the Dimensionless Flow Rate (q/q_max). I suggest using a value different from zero for a default value. Perhaps `1.0`. 
What do you think? Does it make sense?

caio-ramos: Correct, the Vogel's deduction considers a dimensionless flow rate. But, we ended up computing the flow rate as a function of q_max, P_well and P_res, in other words:
```
q = q_max* (1 - 0.2 * (P_well/P_res) - 0.8 * (P_well/P_res)^2)
``` 
So, q_max is no longer in the denominator of a fraction, which means no division by zero.
However, to be consistent with Vogel's approach, using q_max = 1.0 by default is not a bad idea.

src/alfasim_sdk/_internal/alfacase/case_description.py
caio-ramos: I'm not sure about this point. For linear IPR we have a constant value provided by the user to define the slope of the curve. It could be a constant value or transient, leading to a situation where your productivity changes over time, which makes sense as the reservoir pressure will decrease over time, right?

When I was studying Vogel's IPR model, I assumed that q_max could play this same role, which means that it could also vary in time. But I didn't find any reference mentioning it explicitly. Maybe we can allow using just one q_max for the entire simulation to be on the safe side.

src/alfasim_sdk/_internal/alfacase/case_description.py
ggrbill: You are right! I forgot about it. Since we want to have any flow rate in the IPR, I think keeping the default value equal to one  seems to be a good approach.

src/alfasim_sdk/_internal/alfacase/case_description.py
ggrbill: Humm, you are right. We have it to the Linear IPR model. And also there, the curve(linear) changes over time.

It may make sense in a practical way. Let's ask Carol about it, and see what she has in mind.

ggrbill (approved):
Good Job!

After changing the time behavior of the `q_max`, please ask Carol. Thinking twice makes me agree with you about having this feature.

eniohayashi (approved):

prusse-martin:
I did just out of the blue remember where we did document the fix to the problem building the documentation.

https://github.com/ESSS/alfasim-sdk/blob/master/docs/README.rst#alfasim-sdk-documentation

> Remember to add a new case description class to api_reference.rst file. Not doing so will result in errors like this one when generating the docs:
> 
> ```
> source/alfacase_definitions/PositionalPipeTrendDescription.txt:5: WARNING: py:class reference target not found: SurgeVolumeOptionsDescription
> source/alfacase_definitions/PositionalPipeTrendDescription.txt:17: WARNING: py:class reference target not found: SurgeVolumeOptionsDescription
> ```

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1141,6 +1170,9 @@ class IPRModelsDescription:
     linear_models: Dict[str, LinearIPRDescription] = attr.ib(
         default=attr.Factory(dict), validator=dict_of(LinearIPRDescription)
     )
+    vogel_models: Dict[str, VogelIPRDescription] = attr.ib(
+        default=attr.Factory(dict), validator=dict_of(VogelIPRDescription)
+    )
```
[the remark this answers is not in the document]
caio-ramos: I tried, but `inv cog` complains about it.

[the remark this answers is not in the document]
prusse-martin: Probably because we don't use type hints with `attrib_*` guys...
Given this class uses the a manual loader it is OK to directly use `attr.ib` I guess.

prusse-martin (approved):

PR #310, opened 2023-05-22
Minor - Fix efficiencies values

Gualcar:
Reference: ASIM-5218

Efficiency values defined in CaseDescriptions are in fraction, not in percentage.


Gualcar:
This change is related to some bug fixes in the Pump model. Here in `alfasim-sdk` this will only affect tests.

ggrbill:
image

PR #316, opened 2023-07-31
[ASIM-5342] add invoke support

ThalesCarl:
This PR will include the support of ``pyinvoke`` tasks to perform the activities that before were done by the cli interface. This will allow the user of the alfasim-sdk project to use a more extensible toolkit for constructing their plugin, because the user will be able to define their one tasks and overwrite the default tasks. 

Finally, I have updated the readthedocs documentation in the chapters 01 and 03. You can take a loot at https://alfasim-sdk--316.org.readthedocs.build/en/316


ThalesCarl:
Still fixing the CI jobs but I would appreciate the review

src/alfasim_sdk/default_tasks.py
```diff
@@ -0,0 +1,428 @@
...
+
+    if sys.platform == "win32":
+        shared_lib_name = f"{plugin_id}.dll"
+    else:
+        shared_lib_name = f"lib{plugin_id}.so"
+
+    shared_lib_path = artifacts_dir / shared_lib_name
+
+    if build_dir.exists():
+        shutil.rmtree(build_dir)
```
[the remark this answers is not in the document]
ThalesCarl: Great idea. I copied the content from the `compile.py` to the task but I didn't pay attention to this detail. I will include the `--clean` option to the task

src/alfasim_sdk/default_tasks.py
```diff
@@ -0,0 +1,428 @@
...
+    - ´<plugin_folder>/src/hook_specs.h´ |br|
+    """
+    hook_specs_file_path = _get_hook_specs_file_path()
+    hm = HookManGenerator(hook_spec_file_path=hook_specs_file_path)
+
+    # Set to the plugin root, where the tasks.py is located in the plugin structure
+    plugin_folder = Path(ctx.config._project_prefix)
+    hook_specs_h_path = plugin_folder / "src" / "hook_specs.h"
+
+    if not hook_specs_h_path.exists() or not hook_specs_h_path.is_file():
```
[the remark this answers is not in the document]
ThalesCarl: Good catch. I will remove it.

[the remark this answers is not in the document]
prusse-martin: BTW. `.is_file()` will return `False` if the file does not exist.

src/alfasim_sdk/default_tasks.py
```diff
@@ -240,16 +240,21 @@ def install_plugin_to_user_folder(ctx):
...
     except StopIteration:
-        raise RuntimeError(
-            f"Something went wrong on plugin compilation: "
-            f'could not find the "*.hmplugin" in {plugin_folder}.'
+        print_message(
+            f"Could not find any '*.hmplugin' in {plugin_folder}.",
+            color=Fore.RED,
+            bright=True,
         )
+        exit(0)
```
prusse-martin, line 248: Why exit with code `0`?
`0` use to represent no error took place. Is that not an error?

Also, prefer `raise invoke.Exit(code=0)`, or `raise invoke.Exit(msg, code=0)` (`msg` is printed to stderr)

prusse-martin: I think that when using `raise invoke.Exit(code=1)` invoke does not print a stack trace to the console before exiting

src/alfasim_sdk/_internal/cli.py
```diff
@@ -111,6 +122,17 @@ def template(dst, caption, plugin_id, author_name, author_email):
     python_folder.mkdir()
     Path(python_folder / f"{plugin_id}.py").touch()
 
+    if pyinvoke:
+        # remove compile.py created by hookman
+        compile_file = dst / plugin_id / "compile.py"
+        if compile_file.exists():
+            compile_file.unlink()
```
prusse-martin, line 129: Just an idea.

Given hookman create the `compile.py` file. Isn't it better to incorporate part of that file in the tasks somehow?

src/alfasim_sdk/default_tasks.py
ThalesCarl: Good suggestion. 

I didn't put much attention to the code `0`, indeed. I think a different value is a better way to handle it. Which one do you guys suggest?

However, I opted to use a direct `exit()` call instead raising a exception, because I was having some issues with the test of the task. The `invoke` is called using `subprocess.run()` and I even tough I was able to mock the `subprocess.run()` call, the coverage was not marking it as covered.

src/alfasim_sdk/_internal/cli.py
ThalesCarl: I don't know. In the discussion in Zulip's topic, we have stablished that when the user chooses to use invoke, the compilation will be done using just the task. But I think your idea is good to support backwards compatibility. What do you think @nicoddemus?

src/alfasim_sdk/default_tasks.py
ThalesCarl: Just changed the error code to 1.

src/alfasim_sdk/_internal/cli.py
nicoddemus: I not sure I follow what @prusse-martin is suggesting, can you please clarify it?

ThalesCarl: I believe @prusse-martin is suggesting that we still use the `compile.py` file generated, and the task `compile` will just execute `python compile.py` similarly to what the `alfasim-sdk compile` was doing previously. Am I right?

nicoddemus: But what are the benefits of that over just following the pattern of using `inv compile` as for all other tasks? Remember that there's no backward compatibility concerns here.

nicoddemus: AHh OK re-reading I understand the suggestion.

I'm not against the `inv compile` taks just executing `python compile.py` behind the scenes, given `compile.py` is created by hookman.

But does that make sense still, I mean we compile the *plugin* using the `compile.py` unchanged, as (generated by hookman)?

src/alfasim_sdk/_internal/cli.py
```diff
@@ -54,13 +55,23 @@ def console_main():
...
+    help="Configure the new plugin to use pyinvoke tasks instead of the compile.py file",
+)
+def new(dst, pyinvoke, caption, plugin_id, author_name, author_email):
     r"""
     Generate a template with the necessary files and structure to create a plugin.
 
     The template folder will be placed on the ``dst`` option, that by default is the current directory from where the command
     was invoked.
 
+    If the flag ``--pyinvoke`` is provided, the ``compile.py`` will be replaced by the `tasks.py``
```
nicoddemus: I don't think we need this option, from now on our sdk should *always* generate the `tasks.py` file.

src/alfasim_sdk/_internal/cli.py
nicoddemus: See my suggestion, in another comment, I think there's some confusion: we don't need to have two "modes" for new, one legacy and one using "pyinvoke" (with tasks.py) -- `new` should now always just use `tasks.py` and that's it.

ThalesCarl: I think it could give the user the ability to change the compilation steps. For instance, passing a different flag to cmake. But I understand that given we will be using just invoke from now on, the approach would be overwriting the `compile` task instead.

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Ohh I see. I misunderstood this.

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,29 +18,21 @@ the official docs <https://www.python.org/downloads/>`_.
...
-
-Also, make sure to have at least the following version for this software:
+|sdk| requires at least the following version for this software:
 
  - ``CMake`` at least version 3.5.2
  - ``Ninja`` at least version 1.7.0
 
 
-The ``alfasim-sdk`` package has several commands that can be visualized by running the help option.
+The ``alfasim-sdk`` package that you have installed have a built in command line interface (CLI) that is used to create a new plugin template. You can get the
```
eniohayashi, line 34: Minor: `(...) has a built in (...)`

eniohayashi (commented):
Nice work!

eniohayashi (approved):

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,29 +18,21 @@ the official docs <https://www.python.org/downloads/>`_.
...
-    It is recommended that you install the |sdk| using a Python Virtual Environment.
-    For more details, see `Virtual Environments and Packages in Python documentation <https://docs.python.org/3/tutorial/venv.html>`_.
+    It is recommended that you install the |sdk| using the conda package manager.
 
-From a terminal, and inside a virtual environment, update pip:
+    You can it by adding a ``environment.yml`` to your project and declare the ``alfasim-sdk`` inside the ``pip`` dependencies.
+    For more details, see `Getting started with conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`_.
 
-.. code-block:: console
+    And if you want to manage multiple repositories using conda, a further addition to your project could be the `conda devenv <https://github.com/ESSS/conda-devenv>`_ tool.
```
nicoddemus, line 26: I would remove this, probably will generate only noise to most users.

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,29 +18,21 @@ the official docs <https://www.python.org/downloads/>`_.
...
-
-Install the |sdk| with:
-
-.. code-block:: console
-
-    >>> pip install alfasim-sdk
-
-
-Also, make sure to have at least the following version for this software:
+|sdk| requires at least the following version for this software:
```
nicoddemus, line 28:
```suggestion
|sdk| requires at least the following version for these software:
```

Or simply:

```suggestion
|sdk| requires:
```

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,29 +18,21 @@ the official docs <https://www.python.org/downloads/>`_.
...
 
  - ``CMake`` at least version 3.5.2
  - ``Ninja`` at least version 1.7.0
 
 
-The ``alfasim-sdk`` package has several commands that can be visualized by running the help option.
+The ``alfasim-sdk`` package that you have installed have a built in command line interface (CLI) that is used to create a new plugin template. You can get the
+help on how to use it with the following command
```
nicoddemus, line 35:
```suggestion
The ``alfasim-sdk`` package has a built in command line interface (CLI) that is used to create new plugins. You can get help with the following command:
```

docs/source/plugins/01_quick_start.rst
```diff
@@ -104,22 +96,37 @@ The highlights here are for:
 :myplugin.py: Implementation of the hooks for customization of the UI interface, or the solver configuration hooks.
 :myplugin.c:  Implementation of the hooks for customization of solver.
 
-
 Check out the :ref:`Plugin Structure section <plugin_structure-section>` for more details about how the folder and files are structured, and
 also, check the :ref:`plugin-by-example-section` that shows how to create simple plugins that make use of the `User Interface Hooks` and the `Solver Hooks`.
 
+Invoke support
+--------------
```
nicoddemus, line 103: I would not say "invoke support", because "support" conveys the idea that it is somehow "optional", which in fact is central on how we manage tasks now.

```suggestion
Invoke
-------
```

CHANGELOG.rst
```diff
@@ -13,6 +13,7 @@ History
 
 * Removed ``user_defined_esp_table`` and ``esp_parameters`` from ``PumpEquipmentDescription``.
 
+* Added ``pyinvoke`` support.
```
nicoddemus, line 16: "support" conveys the idea that is somehow optional, but is in fact central to how we manage tasks now.

```suggestion
* ``pyinvoke`` is now used to manage plugins tasks.
```

docs/source/plugins/01_quick_start.rst
```diff
@@ -104,22 +96,37 @@ The highlights here are for:
...
+
+From version 0.17.0 onwards, |sdk| have ``pyinvoke`` support in order facilitate and standardize the plugins development process
+using pre-defined tasks. In order to use it, you should create a new plugin with the command below and a ``tasks.py`` file will be generated.
+
+The generetad ``tasks.py`` file can be used to implement your own tasks and to overwrite the default tasks defined in the ``alfasim_sdk.default_tasks`` .
+
+|sdk| already have some tasks available out of the box that will help to speed up your workflow. You might list them using:
```
nicoddemus, line 110:
```suggestion
From version 0.17.0 onwards, |sdk| uses `pyinvoke`_ in order facilitate and standardize the plugins development process
using pre-defined tasks. A standard ``tasks.py`` file is generated automatically by the :program:`new` command, which can be customized by the user if needed.

To list the tasks available, use:
```

docs/source/plugins/01_quick_start.rst
```diff
@@ -104,22 +96,37 @@ The highlights here are for:
...
+
+.. code-block:: bash
+
+    >>> cd myplugin
+    >>> invoke --list
+
+Further details about ``pyinvoke`` can be found on its `documentation <https://www.pyinvoke.org/>`_.
```
nicoddemus, line 117: This will make the `pyinvoke_` link I used in the previous suggestion work:

```suggestion
.. _pyinvoke: https://www.pyinvoke.org
```

docs/source/plugins/03_plugin_structure.rst
```diff
@@ -4,48 +4,41 @@ Plugins Structure
 =================
 
 As mentioned in :ref:`quick-start-section`, the |sdk| package has some utilities to help in the process to create
-a new plugin project and to create a plugin file.
+a plugin file.
 
-At any moment, it is possible to invoke the ``help`` command to list all commands available.
+First, for the creation of the plugin you can use the CLI automatic generator
```
nicoddemus, line 9: We tend in Portuguese to be too verbose when compared to English, I suggest this which is short and to the point:

```suggestion
First, to create the plugin use :program:`new`:
```

docs/source/plugins/03_plugin_structure.rst
```diff
@@ -4,48 +4,41 @@ Plugins Structure
...
-    :prog: alfasim-sdk template
+.. click:: alfasim_sdk._internal.cli:new
+    :prog: alfasim-sdk new
     :show-nested:
 
+Pyinvoke tasks
+--------------
 
-.. _alfasim_sdk_cli_package_section:
+Besides the creation of the plugin struct, all other activities that you might want to perform, you will use ``pyinvoke`` tasks.
```
nicoddemus, line 24:
```suggestion
Except for creating the plugin structure, all other activities you will use ``pyinvoke`` tasks.
```

src/alfasim_sdk/_internal/cli.py
```diff
@@ -54,20 +54,25 @@ def console_main():
     prompt="-- Author Email",
     help="Email of the plugin author, this value is stored in plugin metadata and not displayed on the application explicitly",
 )
-def template(dst, caption, plugin_id, author_name, author_email):
+def new(dst, caption, plugin_id, author_name, author_email):
     r"""
     Generate a template with the necessary files and structure to create a plugin.
```
nicoddemus, line 59:
```suggestion
    Generate a new plugin directory structure and files.
```

src/alfasim_sdk/_internal/cli.py
```diff
@@ -54,20 +54,25 @@ def console_main():
...
 
     The template folder will be placed on the ``dst`` option, that by default is the current directory from where the command
     was invoked.
 
-    The files generated and their contents are ready to be used or customized and have the following structure:
+    A ``tasks.py`` will be generated, where you can define your personalized pyinvoke's tasks.
+    Also, you will have available the default tasks defined at ``src\alfasim-sdk\default_tasks.py``.
+    Run ``inv --list`` to get a list of the available tasks.
```
nicoddemus, line 66:
```suggestion
    A ``tasks.py`` will be generated containing default tasks, and which can be further customized if needed.
    Run ``invoke --list`` to get a list of the available tasks.
```

src/alfasim_sdk/_internal/cli.py
```diff
@@ -54,20 +54,25 @@ def console_main():
...
+
+    The files generated and their contents are ready to be used or customized and have the
+    following structure:
```
nicoddemus, line 69:
```suggestion
    The files generated and their contents are ready to be used and have the
    following structure:
```

src/alfasim_sdk/_internal/cli.py
```diff
@@ -111,121 +116,33 @@ def template(dst, caption, plugin_id, author_name, author_email):
     python_folder.mkdir()
     Path(python_folder / f"{plugin_id}.py").touch()
 
+    # remove compile.py created by hookman
+    compile_file = dst / plugin_id / "compile.py"
+    if compile_file.exists():
```
nicoddemus, line 121:
```suggestion
    if compile_file.is_file():
```

General advice: prefer to be specific if you mean a file or a directory.

src/alfasim_sdk/_internal/cli.py
```diff
@@ -111,121 +116,33 @@ def template(dst, caption, plugin_id, author_name, author_email):
...
-    Generate a ``<package_name>.hmplugin`` file with all the content from the directory assets and artifacts.
+def _invoke_tasks_file_content() -> str:
+    return dedent(
+        """\
+        from alfasim_sdk.default_tasks import *
+        from invoke import task
 
-    By default, the package will be created inside the folder plugin_dir, however, it's possible
-    to give another path filling the dst argument.
+        # Customize this file with your own invoke tasks
```
nicoddemus, line 142:
```suggestion
        # Customize this file with your own invoke tasks.
```

nicoddemus (approved):
Final review of the documentation mostly, as the code itself already has been through extensive reviews.

Leaving pre-approved. 👍 

Great work!

docs/source/plugins/01_quick_start.rst
ThalesCarl: Fixed

docs/source/plugins/01_quick_start.rst
ThalesCarl: Yeah, makes sense. It was removed.

docs/source/plugins/01_quick_start.rst
ThalesCarl: Fixed

docs/source/plugins/01_quick_start.rst
ThalesCarl: What about `Invoke tasks`? Just to have a more meaningful title.

CHANGELOG.rst
ThalesCarl: Fixed

docs/source/plugins/01_quick_start.rst
ThalesCarl: Alright, I've added this suggestion.

docs/source/plugins/01_quick_start.rst
ThalesCarl: Nice.

docs/source/plugins/01_quick_start.rst
ThalesCarl: Fixed

docs/source/plugins/01_quick_start.rst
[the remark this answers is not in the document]
ThalesCarl: Fixed

docs/source/plugins/03_plugin_structure.rst
ThalesCarl: Fixed

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Fixed

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Fixed

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Nice catch, thanks, I've added your suggestion.

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Fixed

docs/source/plugins/03_plugin_structure.rst
```diff
@@ -4,48 +4,41 @@ Plugins Structure
...
 
+.. invoke::
+   :module: alfasim_sdk.default_tasks
+   :prog: invoke
 
-.. click:: alfasim_sdk._internal.cli:package_only
-    :prog: alfasim-sdk package_only
-    :show-nested:
+You can check their implementations `here <https://github.com/ESSS/alfasim-sdk/blob/master/src/alfasim_sdk/default_tasks.py>`_
+and you can also overwrite them by just defining a function with the same name of the default task with the ``@task`` decorator.
```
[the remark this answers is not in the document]
ThalesCarl: Okay, I will leave as it is.

src/alfasim_sdk/default_tasks.py
ThalesCarl: @prusse-martin Can you double check if this is acceptable, please?

src/alfasim_sdk/default_tasks.py
```diff
@@ -0,0 +1,455 @@
...
+
+    try:
+        hm_plugin_path = next(plugin_folder.glob("*.hmplugin"))
+    except StopIteration:
+        print_message(
+            f"Could not find any '*.hmplugin' in {plugin_folder}.",
+            color=Fore.RED,
+            bright=True,
+        )
+        exit(1)
```
nicoddemus, line 270: I think it has been commented before, but although this works, I believe the right way to go about here is to `raise invoke.Exit(code=1)`:

https://docs.pyinvoke.org/en/stable/api/exceptions.html#invoke.exceptions.Exit

src/alfasim_sdk/default_tasks.py
ThalesCarl: Actually, changing to `raise invoke.Exit(code=1)` didn't break my test 😅. So, thanks for the suggestion.

src/alfasim_sdk/_internal/cli.py
prusse-martin: > ... `compile.py` unchanged, as generated by hookman?

This was my point (not backward compatibility).
If the are some relevant change in hookman using `inv compile` will not be affected by it, that is why I am suggesting to make `inv compile` deffer the actual compilation to `compile.py`.



> I think it could give the user the ability to change the compilation steps.

If the author really want that OK. He is totally free to that.

> For instance, passing a different flag to cmake.

This should probably be covered by hookman when generating the `compile.py` file (extra flags, override flags), and the sdk could provide support to configure this using the spec file (we are going to add an spec file, or I am mistaken?).

> But I understand that given we will be using just invoke from now on, the approach would be overwriting the `compile` task instead.

`compile.py` it is a generated file, if the author changes it he is at his own (I maybe wrong and hookman's compile.py be supposed to be changed).
On the `inv compile` front I am not totally sure we want to encourage that. Can think of ppl coming up with a lot of fancy ways of creating incompatible plugins.

(BTW, I did not yet read the rest of the comments, sorry if I get a bit redundant)

docs/source/plugins/03_plugin_structure.rst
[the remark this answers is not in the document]
prusse-martin: "read the docs" will build the docs for your PR, just add a PR comment on the same line with the link from that.

src/alfasim_sdk/_internal/cli.py
ThalesCarl: Currently, the users are already changing the `compile.py`. For instance, check our hydrates plugin (I will not link it here, because this is a public repo).

docs/source/plugins/03_plugin_structure.rst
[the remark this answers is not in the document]
ThalesCarl: I am not sure if I understand what you are suggesting. I was talking that the link to <https://github.com/ESSS/alfasim-sdk/blob/master/src/alfasim_sdk/default_tasks.py> is broken because I am linking a file that still doesnt exist on github at the master branch.

ggrbill (approved):
Great Job. 

Please, consider add a minimal installation instructions.

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,29 +18,20 @@ the official docs <https://www.python.org/downloads/>`_.
...
 
-.. code-block:: console
 
-    >>> python -m pip install -U pip
-
-Install the |sdk| with:
-
-.. code-block:: console
-
-    >>> pip install alfasim-sdk
```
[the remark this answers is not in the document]
ThalesCarl: I've added installation instructions in the line above:

```
 You can it by adding a ``environment.yml`` to your project and declare the ``alfasim-sdk`` inside the ``pip`` dependencies.
```  

but I will rephrase it to be more clearer.

src/alfasim_sdk/_internal/cli.py
ThalesCarl: As discussed in the daily meeting, we can merge it as it is and further improvement could be made afterwards.

docs/source/plugins/03_plugin_structure.rst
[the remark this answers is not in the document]
ThalesCarl: This is a minor thing, I will resolve the conversation, because once merged the link will be valid.

PR #320, opened 2023-09-11
Update installation tasks

ThalesCarl:
This PR will add the `install_dir` parameter to the installation tasks in order to the end user be able to customize where to install the plugins. Also, the tasks are renamed since the installation is no longer performed only for the user folder.

ASIM-4738

### DIFF
--- a/tests/test_invoke.py
+++ b/tests/test_invoke.py
@@ -7,6 +7,7 @@ from typing import List
 from zipfile import ZipFile
 
 import pytest
+from _pytest.monkeypatch import MonkeyPatch
 from ruamel.yaml import YAML
 
 from alfasim_sdk._internal.alfasim_sdk_utils import get_current_version
@@ -45,8 +46,8 @@ def new_plugin_dir(tmp_path: Path) -> Path:
     return plugin_dir
 
 
-def test_compile_task(new_plugin_dir: Path):
-    os.chdir(new_plugin_dir)
+def test_compile_task(new_plugin_dir: Path, monkeypatch: MonkeyPatch):
+    monkeypatch.chdir(new_plugin_dir)
 
     subprocess.run(
         [f"{invoke_cmd}", "compile"],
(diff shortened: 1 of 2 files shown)


ggrbill (approved):
Nice Job!

Is there a need to update plugin tasks? (like Hydrates, Wax, and others)

tests/test_invoke.py
```diff
@@ -192,15 +192,15 @@ def create_fake_hmplugin(plugin_dir: Path) -> Path:
...
     fake_hmplugin_files = get_plugin_installed_files(new_plugin_dir)
 
     user_folder_env_var = "USERPROFILE" if sys.platform == "win32" else "HOME"
     os.environ[user_folder_env_var] = str(tmp_path)
 
-    subprocess.run([f"{invoke_cmd}", "install-plugin-to-user-folder"], env=os.environ)
+    subprocess.run([f"{invoke_cmd}", "install-plugin"], env=os.environ)
 
     install_dir = Path("~/.alfasim_plugins").expanduser()
     os.chdir(install_dir)
```
prusse-martin, line 206: In the future instead doing `os.chdir` and mutating `os.environ` prefer to use the fixture `monkeypatch` and do `monkeypatch.chdir` and `monkeypatch.setenv` and the fixture will restore the original global state affected by changing the cwd or the environment.


Also, when doing:
```
    os.environ[user_folder_env_var] = str(tmp_path)
    subprocess.run([f"{invoke_cmd}", "install-plugin"], env=os.environ)
```
you could do:
```
    env = {**os.environ, user_folder_env_var: str(tmp_path)}
    subprocess.run([f"{invoke_cmd}", "install-plugin"], env=env)
```
The same effect without tainting the global `os.environ` variable.

prusse-martin (approved):

tests/test_invoke.py
ThalesCarl: Thanks for the tip. I've updated the code to use the monkeypatch.

PR #321, opened 2023-09-13
Update readthedocs config to use build.os

### LINKED ISSUE #317: Update readthedocs config to use `build.os` instead of `build.image`
* alfasim-sdk version: 0.17 (unreleased)
* Python version: 3.10
* Operating System: windows

### Description

While working on PR https://github.com/ESSS/alfasim-sdk/pull/316, the readthedocs failed with the following messsage:

```
The configuration key "build.image" is deprecated. Use "build.os" instead to continue building your project. 
Read more at https://docs.readthedocs.io/en/stable/config-file/v2.html#build-os
```

From [this](https://blog.readthedocs.com/use-build-os-config/) readthedocs' announcement, we can see that we need to update the configuration of the project before October 16, 2023.

### What I Did

While working on the PR #316 I have changed the configuration as below
```yaml
version: 2

build:
  os: "ubuntu-22.04"
  tools:
    python: "3.7"

python:
    install:
        -   method: pip
            path: .
        -   requirements: docs/requirements.txt
sphinx:
    configuration: docs/source/conf.py
    fail_on_warning: true
```
It succeed the building, but I am not sure if this is the best approach for the newer version of the readthedocs. Furthermore, the following warning was thrown:

```
WARNING: dot command 'dot' cannot be run (needed for graphviz output), check the graphviz_dot setting
```

Since our policy is to `fail_on_warning`, then this must be correct as well in this issue.

ThalesCarl:
Previously, the config was using build.image, but this configuration will be no longer available after October, 16th 2023. See [this announcement](https://blog.readthedocs.com/use-build-os-config/)

Notice that I've added `graphviz` dependency to the `apt_packages` in the `readthedocs.yml` and removed from the `requirements.txt` because using `pip` the binary seems to not be available anymore, and the `graphviz` documentation suggests that it's preferable to use the package manager.

Finally, I tried changing the python version to be a newer one like 3.10, but there was some problems with the pinned version of sphinx and I opt to leave it to python 3.7 for now.

Close #317
ASIM-5369

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,15 +1,18 @@
 version: 2
 
 build:
-  image: latest
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.10"
+  apt_packages:
+    - graphviz
 
 python:
-    version: 3.7
-    install:
-        -   method: pip
-            path: .
-        -   requirements: docs/requirements.txt
+  install:
+    - method: pip
+      path: .
+    - requirements: docs/requirements.txt
 
 sphinx:
-    configuration: docs/source/conf.py
-    fail_on_warning: true
+  configuration: docs/source/conf.py
+  fail_on_warning: true
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,12 +1,9 @@
-sphinx==4.0.2
-breathe==4.30.0
+sphinx
+breathe
 sphinx-click
 sphinx-invoke
 pydata-sphinx-theme
 sphinx-inline-tabs
 sphinx-copybutton
 sphinx_togglebutton
-graphviz
 colorama
-# Pin docutils because of circular include warnings (see #193)
-docutils==0.16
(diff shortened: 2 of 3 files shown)


.readthedocs.yml
```diff
@@ -1,10 +1,13 @@
 version: 2
 
 build:
-  image: latest
+  os: "ubuntu-22.04"
+  tools:
+    python: "3.7"
```
nicoddemus, line 6: Python 3.7 is EOL, can we update this?

```suggestion
    python: "3.11"
```

nicoddemus (approved):

.readthedocs.yml
nicoddemus: Took the liberty of applying this to see how it fares.

nicoddemus:
Made some changes so we can build the docs using Python 3.10. 👍 

(Make sure to squash all changes)

PR #340, opened 2023-11-08
Add derivatives wrt to enthalpy to PH case description

caio-ramos:
Part of the DCC-93/DCC-94.

Update the constant PVT model available in alfasim-sdk. Essentially, it includes derivatives with respect to enthalpy.
Additionally, the naming standard for derivatives has also been updated.

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/case_description.py
+++ b/src/alfasim_sdk/_internal/alfacase/case_description.py
@@ -2562,19 +2562,19 @@ class PvtModelPtTableParametersDescription:
         def constant_gas_density_model(p, t):
             return rho_g_ref + 0 * p
 
-        def gas_density_derivative_respect_pressure(p, t):
+        def gas_density_derivative_respect_pressure_const_temperature(p, t):
             return 1 / (r * t)
 
-        def gas_density_derivative_respect_temperature(p, t):
+        def gas_density_derivative_respect_temperature_const_pressure(p, t):
             return -p / (r * t**2)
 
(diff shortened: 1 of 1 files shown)


caio-ramos:
Sorry guys, I already had approvals for this branch in another PR (https://github.com/ESSS/alfasim-sdk/pull/323). But as I continued working on the issue, in a moment of distraction I removed the commits locally and ended up doing a push force, I believe that the absence of commits deleted the original PR. I'm reopening it here, the content is essentially what you've seen before.

eniohayashi (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -3043,8 +3084,10 @@ def oil_water_surface_tension_model(p, t):
 
             names += [
                 "water density",
-                "water density derivative pressure",
-                "water density derivative temperature",
+                "water density derivative pressure, const temperature",
+                "water density derivative temperature, const pressure",
+                "water density derivative pressure, constant enthalpy",
+                "water density derivative enthalpy, constant pressure",
```
prusse-martin, line 3090: If we got alfacase files saved with the old names they will be loaded with the old names.
Today we have no support for backward compatibility in alfacase files.
Will this be a problem?

caio-ramos: As we discussed via zulip, I created this issue to investigate this point - ASIM-5482

PR #342, opened 2023-11-29
[ASIM-5490] Add constants to aggregator constanst file

BeneBr:
Adds two constants that contemplates the UQ results captions for their datasets.

[ASIM-5490](https://esss.atlassian.net/browse/ASIM-5490)

[ASIM-5490]: https://esss.atlassian.net/browse/ASIM-5490?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/alfasim_sdk/result_reader/aggregator_constants.py
+++ b/src/alfasim_sdk/result_reader/aggregator_constants.py
@@ -1,5 +1,7 @@
 TRENDS_GROUP_NAME = "trends"
 TRENDS_META_ATTR_NAME = "trends"
+GLOBAL_SENSITIVITY_ANALYSIS_META_ATTR_NAME = "global_sensitivity_analysis"
+GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME = "global_sensitivity_analysis"
 
 TIME_SET_DSET_NAME = "time_set"
 


cauebs (approved):

ro-oliveira95 (approved):

src/alfasim_sdk/result_reader/aggregator_constants.py
```diff
@@ -1,5 +1,7 @@
 TRENDS_GROUP_NAME = "trends"
 TRENDS_META_ATTR_NAME = "trends"
+GLOBAL_SENSITIVITY_ANALYSIS_META_ATTR_NAME = "global_sensitivity_analysis"
+GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME = "global_sensitivity_analysis"
```
prusse-martin, line 4: Won't the actual reading code be added to alfasim-sdk?
Why add just the constants?

prusse-martin (changes requested):

src/alfasim_sdk/result_reader/aggregator_constants.py
BeneBr: Martin i only add this constants because in ```SimFacade``` there are other variables used in there tha was added here, mainly constats that is used to save the results.

PR #343, opened 2023-12-01
Allow user plugins to import from other modules inside it

prusse-martin:
Now user plugins can have a namespace package named
`alfasim_sdk_plugins` (located in the `<plugin_id>/src/python` folder)
inside of which the plugin specific python source could be placed an
later imported.
At runtime alfasim/alfasim-sdk will import a root module also named
`alfasim_sdk_plugins` and manually edit `__path__` so the python's
import machinery look into the user plugins' namespace packages without
the need to further edit `sys.path` and cause possible import name
clashes.

ASIM-5419

### DIFF
--- a/src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
+++ b/src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
@@ -185,7 +185,7 @@ def import_module(path: Path) -> ModuleType:
     import importlib
 
     assert path.is_file()
-    spec = importlib.util.spec_from_file_location("", path)
+    spec = importlib.util.spec_from_file_location(path.stem, path)
     assert spec is not None
     assert spec.loader is not None
     module = importlib.util.module_from_spec(spec)
@@ -197,13 +197,30 @@ def load_plugin_data_structure(plugin_id: str) -> Optional[List[Type]]:
     """
     Obtain the models for a given plugin.
     """
+    import alfasim_sdk_plugins
+
     for candidate in get_plugin_module_candidates(plugin_id):
         if candidate.is_file():
+            # Update the "alfasim_sdk_plugins" namespace.
+            namespace_package_str = None
+            remove_namespace_package = False
+            namespace_package = candidate.parent / "alfasim_sdk_plugins"
+            if namespace_package.is_dir():
+                namespace_package_str = str(namespace_package.absolute())
+                if namespace_package_str not in alfasim_sdk_plugins.__path__:
+                    alfasim_sdk_plugins.__path__.append(namespace_package_str)
+                    remove_namespace_package = True
+
             module = import_module(candidate)
             if hasattr(module, "alfasim_get_data_model_type"):
                 models = module.alfasim_get_data_model_type()
                 if all(hasattr(m, "_alfasim_metadata") for m in models):
                     return models
+
+            # Cleanup "alfasim_sdk_plugins" namespace.
+            # When a valid plugin is found leave the namespace package unchanged.
+            if remove_namespace_package:
+                alfasim_sdk_plugins.__path__.remove(namespace_package_str)
 
     else:
         return None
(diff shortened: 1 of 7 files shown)


tests/alfacase/test_plugin_alfacase_to_case.py
```diff
@@ -43,9 +46,80 @@ def test_get_plugin_module_candidates(monkeypatch):
...
+        importable_plugin: None,
+        importable_plugin_source: Path,
+        datadir: Path,
+        monkeypatch: MonkeyPatch,
+        mocker: MockerFixture,
+    ) -> None:
+        import alfasim_sdk_plugins
+        from alfasim_sdk._internal.alfacase import plugin_alfacase_to_case
+
+        # Create and invalid "importable" plugin.
```
ThalesCarl, line 80: Typo:
```suggestion
        # Create an invalid "importable" plugin.
```

ThalesCarl (approved):

src/alfasim_sdk/_internal/alfacase/plugin_alfacase_to_case.py
```diff
@@ -197,14 +197,31 @@ def load_plugin_data_structure(plugin_id: str) -> Optional[List[Type]]:
...
             if hasattr(module, "alfasim_get_data_model_type"):
                 models = module.alfasim_get_data_model_type()
                 if all(hasattr(m, "_alfasim_metadata") for m in models):
                     return models
 
+            # Cleanup "alfasim_sdk_plugins" namespace.
+            # When a valid plugin is found leave the namespace package unchanged.
+            if remove_namespace_package:
+                alfasim_sdk_plugins.__path__.remove(namespace_package_str)
+
```
ro-oliveira95, line 224: Just to confirm: Is it expected to have both conditions satisfied, and If so it's ok to not clean the "alfasim_sdk_plugins" namespace in this case?

prusse-martin: Yes:

- already is there: when alfasim is exporting an `.alfacase` this code will be triggered to export the relevant user plugin data. In this case the user plugin's package will already be accessible and should remain available after;
- don't cleanup: if the returned module does some lazy imports (local to a function) that import will fail if we cleanup after finding the plugin spec module;
- why check before add: I don't like the idea of having multiple equal entries, or invalid entries;

BeneBr (approved):

ro-oliveira95 (approved):

nicoddemus (changes requested):
Can you please update the PR description and commit with the approach taken here and how this solves the original issue?

Can you also update the docs?

prusse-martin:
> Can you please update the PR description and commit with the approach taken here and how this solves the original issue?

I did rebase to add an description of that is being done (the same as the pr description now).

> Can you also update the docs?

I edited the "new" command documentation to mention the new feature, that is added to the docs.
Should I add a whole section in the docs to describe this?

https://alfasim-sdk--343.org.readthedocs.build/en/343/plugins/03_plugin_structure.html#alfasim-sdk-new

src/alfasim_sdk/_internal/cli.py
```diff
@@ -68,22 +68,28 @@ def new(dst, caption, plugin_id, author_name, author_email):
...
+                |   <plugin_id>.cpp
+                |
+                \---python
+                    |   <plugin_id>.py
+                    |
+                    \---alfasim_sdk_plugins
+                        \---<plugin_id>
+                                __init__.py
+
+    Any python code placed in `alfasim_sdk_plugins/<plugin_id>` is importable by `import alfasim_sdk_plugins.<plugin_id>.<my_extra_module>`.
```
nicoddemus, line 92: Great!

nicoddemus (approved):

src/alfasim_sdk/_internal/cli.py
ThalesCarl, line 92: Just a minor change to improve the documentation:
```suggestion
    Any python code placed in ``alfasim_sdk_plugins/<plugin_id>`` is importable by using ``import alfasim_sdk_plugins.<plugin_id>.<my_extra_module>``.
```

prusse-martin: let's see how sphinx hanble this...

PR #344, opened 2023-12-05
Set correct category for void_fractions array

cauebs:
ASIM-5451

This fixes the issue with exporting and then importing an alfacase containing a pump.

I don't quite know how to test this. I haven't found a simple way to reproduce the problem (that doesn't involve launching the application, exporting and then importing the alfacase, then clicking the pump).


nicoddemus:
> I don't quite know how to test this.

In this case I don't think we need a specific test, we are fixing an incorrect definition.

nicoddemus (approved):

ro-oliveira95 (approved):

PR #346, opened 2023-12-21
[ASIM-5231] Add field enable_fast_compositional to NumericalOptionsDescription

ro-oliveira95:
ASIM-5231


nicoddemus (approved):
Please also add a new entry to the CHANGELOG.

BeneBr (approved):

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -707,6 +707,7 @@
     caching_rtol=3e-3,
     caching_atol=4e-4,
     always_repeat_timestep=False,
+    enable_fast_compositional=True,
```
prusse-martin, line 710: Just a comment.

I prefer to use non default values for those test inputs when possible.
The "IO" test will probably pass even if the value is not being written/read right.

PR #348, opened 2023-12-26
WIP

ThalesCarl:


requirements_dev.txt
```diff
@@ -1,3 +1,4 @@
+catch2
```
prusse-martin, line 1: https://en.wikipedia.org/wiki/Catch-22

PR #349, opened 2023-12-28
Update CHANGELOG about release 0.19.0

ro-oliveira95:

PR #351, opened 2023-12-28
Bump to version 0.20.0

ro-oliveira95:
As dictated by our release process, bumping to 0.20 after the 0.19 release.


ThalesCarl (approved):

nicoddemus:
Btw it would have been fine to do the same as #349 in this PR (in a separate commit).

nicoddemus (approved):

CHANGELOG.rst
prusse-martin, line 1: 2 UNRELEASED?

nicoddemus: Good catch, whoever did the `0.19.0` release forgot to update the date.

nicoddemus: Nevermind, it is right on `master`: https://github.com/ESSS/alfasim-sdk/blob/master/CHANGELOG.rst

ro-oliveira95: Actually the 2 unreleased sessions was intentional, the idea was that the 1.19.0 would be filled in #349. But I just realized this was a bad approach, because now branch rb-ASIM-v23.2.0 is conflicting with master in this exact line... I'm fixing it in #355.

PR #357, opened 2024-01-18
Read Global Sensitivity Analysis results

BeneBr:
This fix enable to read the global sensitivity analysis from their file result.

ASIM-5516

### DIFF
--- a/src/alfasim_sdk/result_reader/aggregator.py
+++ b/src/alfasim_sdk/result_reader/aggregator.py
@@ -1,9 +1,10 @@
 import functools
+import json
 import os
 from collections import namedtuple
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any
+from typing import Any, Iterator
 from typing import Callable
 from typing import DefaultDict
 from typing import Dict
@@ -14,6 +15,7 @@ from typing import Tuple
 import attr
 import h5py
 import numpy
+import numpy as np
 
 from alfasim_sdk.result_reader.aggregator_constants import META_GROUP_NAME
 from alfasim_sdk.result_reader.aggregator_constants import PROFILES_GROUP_NAME
@@ -24,6 +26,8 @@ from alfasim_sdk.result_reader.aggregator_constants import RESULT_FILE_LOCKING_M
 from alfasim_sdk.result_reader.aggregator_constants import RESULT_FILE_PREFIX
 from alfasim_sdk.result_reader.aggregator_constants import TIME_SET_DSET_NAME
 from alfasim_sdk.result_reader.aggregator_constants import TRENDS_GROUP_NAME
+from alfasim_sdk.result_reader.aggregator_constants import GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+from typing_extensions import Self
 
 OutputKeyType = str
 """\
@@ -84,6 +88,101 @@ class ResultsNeedFullReloadError(RuntimeError):
     "restart file" and the option "keep old results" is selected.
     """
 
+@attr.s(slots=True, hash=False)
+class GlobalSensitivityAnalysisMetadata:
+    """
+    A class that hold the global sensitivity analysis metadata.
+    """
+
+    @attr.s(slots=True, hash=False)
+    class GSAItem:
+        """
+        A class that hold each global sensitivity analysis
+        metadata information.
+        :ivar property_id:
+            The property name (holdup, total mass flow rate).
+        :ivar trend_id:
+            The id associated a specific trend.
+        :ivar category:
+            The property category. For global sensitivity analysis
+            this always be dimensionless.
+        :ivar parametric_var_id:
+            The id associated a parametric variable.
+        :ivar parametric_var_name:
+            The name associated a parametric variable.
+        :ivar network_element_name:
+            The name of network where this data is associated.
+        :ivar position:
+            The element position in a pipe.
+        :ivar position_unit:
+            The position unit (m, Km, cm).
+        :ivar unit:
+            The unit of a dimensionless property (-, %)
+        :ivar qoi_index:
+            The index of a quantity of interest in the results file.
+        :ivar qoi_data_index:
+            The data index of a quantity of interest.
+        """
+        property_id: str = attr.ib(validator=attr.validators.instance_of(str))
+        trend_id: str = attr.ib(validator=attr.validators.instance_of(str))
+        category: str = attr.ib(validator=attr.validators.instance_of(str))
+        parametric_var_id: str = attr.ib(validator=attr.validators.instance_of(str))
+        parametric_var_name: str = attr.ib(validator=attr.validators.instance_of(str))
+        network_element_name: Optional[str] = attr.ib(
+            validator=attr.validators.optional(attr.validators.instance_of(str)))
+        position: Optional[float] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(float)))
+        position_unit: Optional[str] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(str)))
+        unit: str = attr.ib(validator=attr.validators.instance_of(str))
+        qoi_index: Optional[int] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
+        qoi_data_index: Optional[int] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
+
+        @classmethod
+        def from_dict(cls, data: Dict[str, Any]) -> Self:
+            return cls(
+                property_id=data['property_id'],
+                trend_id=data['trend_id'],
+                category=data['category'],
+                parametric_var_id=data['parametric_var_id'],
+                parametric_var_name=data['parametric_var_name'],
+                network_element_name=data['network_element_name'],
+                position=data['position'],
+                position_unit=data['position_unit'],
+                unit=data['unit'],
+                qoi_index=data['qoi_index'],
+                qoi_data_index=data['qoi_data_index']
+            )
+    gsa_items: Dict[str, GSAItem] = attr.ib(validator=attr.validators.instance_of(Dict))
+    result_directory: Path = attr.ib(validator=attr.validators.instance_of(Path))
+
+    @classmethod
+    def empty(cls, result_directory: Path) -> Self:
+        return GlobalSensitivityAnalysisMetadata(
+            gsa_items={},
+            result_directory=result_directory
+        )
+    @classmethod
+    def get_metadata_from_dir(cls, result_directory: Path) -> Self:
+        """
+        Return the metadata info from result directory.
+        If the directory does not exist/is invalid, return an empty metadata.
+        :param result_directory:
+            The directory result.
+        """
+        def map_data(gsa_metadata: Dict) -> Dict[str, GlobalSensitivityAnalysisMetadata.GSAItem]:
+            return {
+                key: GlobalSensitivityAnalysisMetadata.GSAItem.from_dict(data)
+                for key, data in gsa_metadata.items()
+            }
+
+        with open_global_sensitivity_analysis_result_file(result_directory=result_directory) as result_file:
+            if not result_file:
+                return cls.empty(result_directory=result_directory)
+
+            loaded_metadata = json.loads(result_file[META_GROUP_NAME].attrs["global_sensitivity_analysis"])
+            return cls(
+                gsa_items=map_data(loaded_metadata),
+                result_directory=result_directory
+            )
 
 @attr.s(slots=True, hash=False)
 class ALFASimResultMetadata:
@@ -179,42 +278,6 @@ def open_result_files(result_directory: Path) -> Dict[int, h5py.File]:
 
     Note that once the container dict is collected the files originally returned are closed.
     """
-
-    h5py_file = h5py.File
-    if h5py.version.version_tuple[:2] >= (3, 5):
-        h5py_file = functools.partial(h5py.File, locking=RESULT_FILE_LOCKING_MODE)
-
-    def open_result_file(filename: Path) -> h5py.File:
-        # The lib h5py will fail to open the file for some exotic paths,
-        # but will work if the cwd are such paths.
-        directory = filename.parent
-        old_directory = os.getcwd()
-        try:
-            os.chdir(directory)
-            try:
-                return h5py_file(filename.name, "r", libver="latest", swmr=True)
-            except OSError as os_error:
-                swmr_message = (
-                    "Unable to open file (file is not already open for SWMR writing)"
-                )
-                if str(os_error) == swmr_message:
-                    return h5py_file(filename.name, "r", libver="latest", swmr=False)
-                raise
-
-        except PermissionError:
-            raise PermissionError(
-                f"Could not access folder {filename}.\n"
-                f'If you are using your local "Downloads" directory, consider changing the project file\n'
-                f"to somewhere else as some services try to synchronize that directory and interfere with\n"
-                f"the simulation.\n"
-                "\n"
-                f'1 - Click in "Save as".\n'
-                f"2 - Select a different path, outside Download folder.\n"
-                f"3 - Run simulation again.\n"
-            )
-        finally:
-            os.chdir(old_directory)
-
     # When a new result file is created its metadata contents are not complete, and the file
     # has not been put into SWMR mode yet (SWMR mode does not allow new groups, attributes,
     # and/or data sets to be created).
@@ -243,7 +306,7 @@ def open_result_files(result_directory: Path) -> Dict[int, h5py.File]:
     prefix_len = len(RESULT_FILE_PREFIX)
     result_files.difference_update(files_under_creation)  # Ignore incomplete files.
     result_files_sorted = (
-        (int(filename.name[prefix_len:]), open_result_file(filename))
+        (int(filename.name[prefix_len:]), _open_result_file(filename))
         for filename in result_files
     )
     result_files_sorted_dict = dict(sorted(result_files_sorted, key=lambda x: x[0]))
@@ -254,6 +317,41 @@ def open_result_files(result_directory: Path) -> Dict[int, h5py.File]:
         for f in result_files_sorted_dict.values():
             f.close()
 
+def _open_result_file(filename: Path) -> h5py.File:
+
+    h5py_file = h5py.File
+    if h5py.version.version_tuple[:2] >= (3, 5):
+        h5py_file = functools.partial(h5py.File, locking=RESULT_FILE_LOCKING_MODE)
+
+    # The lib h5py will fail to open the file for some exotic paths,
+    # but will work if the cwd are such paths.
+    directory = filename.parent
+    old_directory = os.getcwd()
+    try:
+        os.chdir(directory)
+        try:
+            return h5py_file(filename.name, "r", libver="latest", swmr=True)
+        except OSError as os_error:
+            swmr_message = (
+                "Unable to open file (file is not already open for SWMR writing)"
+            )
+            if str(os_error) == swmr_message:
+                return h5py_file(filename.name, "r", libver="latest", swmr=False)
+            raise
+
+    except PermissionError:
+        raise PermissionError(
+            f"Could not access folder {filename}.\n"
+            f'If you are using your local "Downloads" directory, consider changing the project file\n'
+            f"to somewhere else as some services try to synchronize that directory and interfere with\n"
+            f"the simulation.\n"
+            "\n"
+            f'1 - Click in "Save as".\n'
+            f"2 - Select a different path, outside Download folder.\n"
+            f"3 - Run simulation again.\n"
+        )
+    finally:
+        os.chdir(old_directory)
 
 def _get_number_of_base_time_steps_from_time_set_info(
     time_set_info_source_dict: Dict,
@@ -1433,3 +1531,53 @@ def concatenate_metadata(
         result_directory=r_a.result_directory,
         app_version_info=app_version_info,
     )
+
+@contextmanager
+def open_global_sensitivity_analysis_result_file(result_directory: Path) -> Iterator[Optional[h5py.File]]:
+    """
+    Open a global sensitivity analysis result file.
+    :param result_directory:
+        The path to result directory.
+    """
+    filename = result_directory / "uq_result"
+    ignored_file = result_directory / "uq_result.creating"
+
+    if not filename.is_file():
+        yield None
+    # Avoid to read result files with incomplete metadata.
+    elif not ignored_file.is_file():
+        with _open_result_file(filename) as file:
+            yield file
+    else:
+        yield None
+def read_global_sensitivity_analysis_meta_data(result_directory: Path) -> Optional[GlobalSensitivityAnalysisMetadata]:
+    """
+    Read the global sensitivity analysis metadata persisted in a result file.
+    """
+
+    return GlobalSensitivityAnalysisMetadata.get_metadata_from_dir(result_directory=result_directory)
+
+def read_global_sensitivity_analysis_time_set(result_directory: Path) -> Optional[numpy.array]:
+    """
+    Get the time set for sensitivity analysis results.
+    """
+    with open_global_sensitivity_analysis_result_file(result_directory=result_directory) as result_file:
+        if not result_file:
+            return
+        return result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]['time_set'][:]
+
+def read_global_sensitivity_coefficients(
+    coefficients_key: str,
+    metadata: GlobalSensitivityAnalysisMetadata,
+) -> Optional[np.ndarray]:
+    """
+    Read the global sensitivity analysis coefficients results.
+    """
+    # The metadata is empty.
+    if not metadata.gsa_items:
+        return None
+    meta = metadata.gsa_items[coefficients_key]
+    with open_global_sensitivity_analysis_result_file(metadata.result_directory) as result_file:
+        gsa_group = result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]
+        coefficients_dset = gsa_group["global_sensitivity_analysis"]
+        return coefficients_dset[meta.qoi_index, meta.qoi_data_index]
--- a/src/alfasim_sdk/result_reader/aggregator.py
+++ b/src/alfasim_sdk/result_reader/aggregator.py
@@ -4,10 +4,11 @@ import os
 from collections import namedtuple
 from contextlib import contextmanager
 from pathlib import Path
-from typing import Any, Iterator
+from typing import Any
 from typing import Callable
 from typing import DefaultDict
 from typing import Dict
+from typing import Iterator
 from typing import List
 from typing import Optional
 from typing import Tuple
@@ -16,7 +17,11 @@ import attr
 import h5py
 import numpy
 import numpy as np
+from typing_extensions import Self
 
+from alfasim_sdk.result_reader.aggregator_constants import (
+    GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME,
+)
 from alfasim_sdk.result_reader.aggregator_constants import META_GROUP_NAME
 from alfasim_sdk.result_reader.aggregator_constants import PROFILES_GROUP_NAME
 from alfasim_sdk.result_reader.aggregator_constants import (
@@ -26,8 +31,6 @@ from alfasim_sdk.result_reader.aggregator_constants import RESULT_FILE_LOCKING_M
 from alfasim_sdk.result_reader.aggregator_constants import RESULT_FILE_PREFIX
 from alfasim_sdk.result_reader.aggregator_constants import TIME_SET_DSET_NAME
 from alfasim_sdk.result_reader.aggregator_constants import TRENDS_GROUP_NAME
-from alfasim_sdk.result_reader.aggregator_constants import GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
-from typing_extensions import Self
 
 OutputKeyType = str
 """\
@@ -88,6 +91,7 @@ class ResultsNeedFullReloadError(RuntimeError):
     "restart file" and the option "keep old results" is selected.
     """
 
+
 @attr.s(slots=True, hash=False)
 class GlobalSensitivityAnalysisMetadata:
     """
@@ -123,43 +127,54 @@ class GlobalSensitivityAnalysisMetadata:
         :ivar qoi_data_index:
             The data index of a quantity of interest.
         """
+
         property_id: str = attr.ib(validator=attr.validators.instance_of(str))
         trend_id: str = attr.ib(validator=attr.validators.instance_of(str))
         category: str = attr.ib(validator=attr.validators.instance_of(str))
         parametric_var_id: str = attr.ib(validator=attr.validators.instance_of(str))
         parametric_var_name: str = attr.ib(validator=attr.validators.instance_of(str))
         network_element_name: Optional[str] = attr.ib(
-            validator=attr.validators.optional(attr.validators.instance_of(str)))
-        position: Optional[float] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(float)))
-        position_unit: Optional[str] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(str)))
+            validator=attr.validators.optional(attr.validators.instance_of(str))
+        )
+        position: Optional[float] = attr.ib(
+            validator=attr.validators.optional(attr.validators.instance_of(float))
+        )
+        position_unit: Optional[str] = attr.ib(
+            validator=attr.validators.optional(attr.validators.instance_of(str))
+        )
         unit: str = attr.ib(validator=attr.validators.instance_of(str))
-        qoi_index: Optional[int] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
-        qoi_data_index: Optional[int] = attr.ib(validator=attr.validators.optional(attr.validators.instance_of(int)))
+        qoi_index: Optional[int] = attr.ib(
+            validator=attr.validators.optional(attr.validators.instance_of(int))
+        )
+        qoi_data_index: Optional[int] = attr.ib(
+            validator=attr.validators.optional(attr.validators.instance_of(int))
+        )
 
         @classmethod
         def from_dict(cls, data: Dict[str, Any]) -> Self:
             return cls(
-                property_id=data['property_id'],
-                trend_id=data['trend_id'],
-                category=data['category'],
-                parametric_var_id=data['parametric_var_id'],
-                parametric_var_name=data['parametric_var_name'],
-                network_element_name=data['network_element_name'],
-                position=data['position'],
-                position_unit=data['position_unit'],
-                unit=data['unit'],
-                qoi_index=data['qoi_index'],
-                qoi_data_index=data['qoi_data_index']
+                property_id=data["property_id"],
+                trend_id=data["trend_id"],
+                category=data["category"],
+                parametric_var_id=data["parametric_var_id"],
+                parametric_var_name=data["parametric_var_name"],
+                network_element_name=data["network_element_name"],
+                position=data["position"],
+                position_unit=data["position_unit"],
+                unit=data["unit"],
+                qoi_index=data["qoi_index"],
+                qoi_data_index=data["qoi_data_index"],
             )
+
     gsa_items: Dict[str, GSAItem] = attr.ib(validator=attr.validators.instance_of(Dict))
     result_directory: Path = attr.ib(validator=attr.validators.instance_of(Path))
 
     @classmethod
     def empty(cls, result_directory: Path) -> Self:
         return GlobalSensitivityAnalysisMetadata(
-            gsa_items={},
-            result_directory=result_directory
+            gsa_items={}, result_directory=result_directory
         )
+
     @classmethod
     def get_metadata_from_dir(cls, result_directory: Path) -> Self:
         """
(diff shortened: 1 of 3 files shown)


src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -255,6 +234,38 @@ def open_result_file(filename: Path) -> h5py.File:
             f.close()
 
 
+def _open_result_file(filename: Path, h5py_file: h5py.File) -> h5py.File:
```
nicoddemus, line 237: - [ ] This `h5py_file` parameter does not make sense, as it will always be the same object (either `h5py.File` or a `partial` only depending on the current `h5py` version), remove the parameter and inline that code inside this function.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1444,87 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+def get_empty_global_sensitivity_analysis_result_metadata(
+    result_directory: Path,
+) -> UQResultMetadata:
+    """
+    Obtain an empty global sensitivity analysis meta-data.
+    """
+
+    return UQResultMetadata(
+        global_gsa_coefficients={}, result_directory=result_directory
+    )
```
nicoddemus, line 1470: - [ ] Instead of this function, create a classmethod `UQResultMetadata.empty()`.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1444,87 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    else:
+        result_file = open_global_sensitivity_analysis_result_file(
+            results_directory=result_directory
+        )
+        return UQResultMetadata(
+            global_gsa_coefficients=json.loads(
+                result_file[META_GROUP_NAME].attrs["global_sensitivity_analysis"]
+            ),
+            result_directory=result_directory,
+        )
```
nicoddemus, line 1498: Also turn this into a classmethod: `UQResultMetadata.from_file()`.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1444,87 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+
+    if not result_directory.is_dir():
+        return get_empty_global_sensitivity_analysis_result_metadata(
+            result_directory=result_directory,
+        )
+
+    if not Path(result_directory / "uq_result").is_file():
+        return get_empty_global_sensitivity_analysis_result_metadata(
+            result_directory=result_directory,
+        )
```
nicoddemus, line 1488: This is questionable... if a file does not exist, it will return an empty metadata instead. We  don't see that usually, usually we get an error instead. Why did you decide on this design?

If it was not for a very good reason, we should probably behave like everybody else, and let the caller check if the file exists or not in order to decide what to do.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1444,87 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+def read_read_global_sensitivity_analysis_time_set(
+    result_directory: Path,
+) -> numpy.array:
+    """
+    Get the time set for sensitivity analysis results.
+    """
+    result_file = open_global_sensitivity_analysis_result_file(
+        results_directory=result_directory
+    )
+    return result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]["time_set"][:]
```
nicoddemus, line 1510: - [ ] Shouldn't you close the file?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1444,87 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+
+
+def read_global_sensitivity_coefficients(
+    results_directory: Path,
+) -> Dict[str, numpy.array]:
+    """
+    Read the global sensitivity analysis coefficients results.
+    """
+    uq_meta_data = read_global_sensitivity_analysis_meta_data(results_directory)
+    result_file = open_global_sensitivity_analysis_result_file(results_directory)
```
nicoddemus, line 1520: - [ ] Ditto here about closing this file.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +89,12 @@ class ResultsNeedFullReloadError(RuntimeError):
     """
 
 
+@attr.s(slots=True, hash=False)
+class UQResultMetadata:
+    global_gsa_coefficients = attr.ib(validator=attr.validators.instance_of(dict))
+    result_directory = attr.ib(validator=attr.validators.instance_of(Path))
```
nicoddemus, line 95: - [ ] Type annotations please:

```suggestion
@attr.s(slots=True, hash=False)
class UQResultMetadata:
    global_gsa_coefficients: dict[???, ???] = attr.ib(validator=attr.validators.instance_of(dict))
    result_directory: Path = attr.ib(validator=attr.validators.instance_of(Path))
```

nicoddemus (changes requested):

prusse-martin (changes requested):

tests/conftest.py
```diff
@@ -173,3 +180,76 @@ def importable_plugin(
...
+            "unit": "degC",
+            "qoi_index": 0,
+            "qoi_data_index": 1,
+        },
+    }
+
+    meta_group.attrs["global_sensitivity_analysis"] = json.dumps(fake_meta)
+    time_set_dset = global_sensitivity_analysis_group[TIME_SET_DSET_NAME]
+    new_time_set_shape = time_set.shape
+    time_set_dset.resize(new_time_set_shape)
```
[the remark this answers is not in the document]
BeneBr: They dont really dont, my bad.

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin: Raising will require the generator source to handle that just to behave as having received an empty metadata.

nicoddemus: A fair trade for a "sane" API IMHO.

tests/conftest.py
```diff
@@ -173,3 +180,79 @@ def importable_plugin(
...
+    time_set_dset[:] = time_set
+    gsa_data_set = global_sensitivity_analysis_group[
+        GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+    ]
+    gsa_data_set[:] = global_sensitivity_analysis
+    return result_dir
```
prusse-martin, line 258:
```suggestion
    gsa_data_set[:] = global_sensitivity_analysis
    file.close()
    return result_dir
```

prusse-martin (changes requested):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +89,49 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+    @classmethod
+    def empty(cls, result_directory: Path) -> Optional["UQResultMetadata"]:
+        return UQResultMetadata(
+            global_gsa_meta_data={}, result_directory=result_directory
+        )
+
+    @classmethod
+    def get_meta_data_from_file(
+        cls, result_directory: Path
+    ) -> Optional["UQResultMetadata"]:
```
nicoddemus, line 108:
```suggestion
    def get_meta_data_from_dir(
        cls, result_directory: Path
    ) -> Optional["UQResultMetadata"]:
```

The name of the method should match the parameter I think

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +89,49 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+        Return the metadata info from result file.
+        :param result_directory:
+            The directory result.
+        """
+        if not result_directory.is_dir():
+            return cls.empty(result_directory=result_directory)
+
+        if not Path(result_directory / "uq_result").is_file():
+            return None
+        else:
```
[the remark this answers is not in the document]
nicoddemus: Agree this this API as is is wonky.

I said previously that we might want to raise an error in case the path does not exist, but I'm not against doing like Martin is suggesting:

If the directory or file does not exist, return an empty metadata, otherwise read and return it.

ro-oliveira95 (changes requested):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1480,64 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+def read_read_global_sensitivity_analysis_time_set(
+    result_directory: Path,
+) -> numpy.array:
+    """
+    Get the time set for sensitivity analysis results.
+    """
+    with open_global_sensitivity_analysis_result_file(
+        results_directory=result_directory
+    ) as result_file:
+        return result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]["time_set"][:]
```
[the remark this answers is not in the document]
BeneBr: I belive that not occur. When a ```h5py.File``` file is created you specify what is the groups and which data sets will be within this group. Even whe have a file without anything inside them, they will respect their specification order.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1480,64 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    """
+    uq_meta_data = read_global_sensitivity_analysis_meta_data(results_directory)
+    with open_global_sensitivity_analysis_result_file(results_directory) as result_file:
+        global_sensitivity_analysis_coefficients = result_file[
+            GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+        ]["global_sensitivity_analysis"][:]
+
+    coefficients_key = f"{property_id}::{parametric_var_id}@{trend_id}"
+
+    coefficient_meta_data = uq_meta_data.global_gsa_meta_data[coefficients_key]
```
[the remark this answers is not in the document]
BeneBr: If the file exists, they always have this key. As you can se in the ```SimFacade.SubmitUQResults```.
```python
SubmitUQResults(self, uq_results: UQSimulationResults) -> None:
        """
        Submit the uq results to a results file.
        :param uq_results:
            The uq results data.
        """
        import json

        import numpy

        from alfasim_sdk.result_reader.aggregator_constants import (
            GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME,
            META_GROUP_NAME,
            RESULTS_FOLDER_NAME,
            TIME_SET_DSET_NAME,
        )

        require_metadata_update = uq_results.require_metadata_update
        if require_metadata_update or (self._current_file is None):
            self.CloseResultFile()
            results_folder = self._directory / RESULTS_FOLDER_NAME
            results_folder.mkdir(parents=True, exist_ok=True)
            filename = 'uq_result'
            ignore_file = results_folder / (filename + '.creating')
            ignore_file.touch()
            self._CreateResultsFile(results_folder=results_folder, filename=filename)

            require_metadata_update = True

        assert self._current_file is not None
        f = self._current_file
        if require_metadata_update:
            meta_group = self._CreateMetaGroup(f)
            global_sensitivity_analysis_group = f.create_group(
                GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME, track_order=True
            )
            uq_meta_result_as_dict = {}
            for qoi_index, qoi_item in enumerate(uq_results.gsa_metadata):
                for result_index, result in enumerate(qoi_item):
                    new_key = (
                        f"{result['property']}::{result['parametric_var_id']}@{result['trend_id']}"
                    )
                    result['qoi_index'] = qoi_index
                    result['qoi_data_index'] = result_index
                    uq_meta_result_as_dict[new_key] = result

            meta_group.attrs['global_sensitivity_analysis'] = json.dumps(uq_meta_result_as_dict)
            global_sensitivity_analysis_group.create_dataset(
                TIME_SET_DSET_NAME, shape=(0,), dtype=numpy.float64, maxshape=(None,)
            )
            global_sensitivity_analysis_group.create_dataset(
                GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME,
                shape=(0, 0, 0),
                dtype=numpy.float64,
                maxshape=(None, None, None),
            )

            uq_results.require_metadata_update = False
            f.flush()
            f.swmr_mode = True
            ignore_file.unlink()

        else:
            meta_group = f[META_GROUP_NAME]
            global_sensitivity_analysis_group = f[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]

        if len(uq_results.global_sensitivity_coefficients) > 0:
            # GlobalSensitivityAnalysis / timeset ---------------------------------------------------------------------
            time_set_dset = global_sensitivity_analysis_group[TIME_SET_DSET_NAME]
            new_time_set_shape = uq_results.time_set.shape
            time_set_dset.resize(new_time_set_shape)
            time_set_dset[:] = uq_results.time_set
            # GlobalSensitivityAnalysis / GlobalSensitivityAnalysis ---------------------------------------------------
            gsa_data_set = global_sensitivity_analysis_group[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]
            gsa_data_set.resize(uq_results.global_sensitivity_coefficients.shape)
            gsa_data_set[:] = uq_results.global_sensitivity_coefficients

        f.flush()
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +90,49 @@ class ResultsNeedFullReloadError(RuntimeError):
     """
 
 
+@attr.s(slots=True, hash=False)
+class UQResultMetadata:
+    global_gsa_meta_data: Dict[str, Any] = attr.ib(
+        validator=attr.validators.instance_of(dict)
+    )
```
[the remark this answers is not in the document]
BeneBr: Excelent approac my dear. I update wiht this. Thanks.

src/alfasim_sdk/result_reader/aggregator.py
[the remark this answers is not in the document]
ro-oliveira95: I see, but note that the variables that compose the `coefficient_key` string are all received as arguments, so a client of this function could easily pass something that doesn't match any key, e.g. with typos on `property_id` or simply because the caller expects some key on the results that for any reason doesn't exist.

But again, this would be a more cautious approach, to handle border cases that in fact could never blow up, so it's up to you. Personally I like to take a more "defensive" approach, to cope with weird scenarios (or at least making then more visible) beforehand.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1513,72 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    """
+    Open a global sensitivity analysis result file.
+    """
+    filename = results_directory / "uq_result"
+    ignored_file = results_directory / "uq_result.creating"
+    # Avoid to read result files with incomplete metadata.
+    if not ignored_file.is_file():
+        file = _open_result_file(filename)
+        yield file
+        file.close()
```
nicoddemus, line 1531: Always surround with try/finally, otherwise the code after the `yield` will not be reached in the case of an exception:

```suggestion
        file = _open_result_file(filename)
        try:
            yield file
        finally:    
            file.close()
```

However, `h5py.File` supports the context manager, so this is even better:

```suggestion
        with _open_result_file(filename):
            yield file    
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1513,72 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+
+    coefficients_key = f"{property_id}::{parametric_var_id}@{trend_id}"
+    meta_information = uq_meta_data.global_sensitivity_analysis_meta_data[
+        coefficients_key
+    ]
+    return {
+        coefficients_key: global_sensitivity_analysis_coefficients[
+            meta_information.qoi_index, meta_information.qoi_data_index
+        ]
+    }
```
ro-oliveira95, line 1584: Why returning a dict if it has only one key? Maybe a tuple with the coefficient and the results would be more intuitive for users of this function.
```suggestion
    return (
        coefficients_key, 
        global_sensitivity_analysis_coefficients[
            meta_information.qoi_index, meta_information.qoi_data_index
        ]
    )
```

ro-oliveira95 (changes requested):
Requesting change only for the wrongly defined type annotation.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1513,72 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    """
+    Read the global sensitivity analysis metadata persisted in a result file.
+    """
+
+    return GlobalSensitivityAnalysisMetaData.get_meta_data_from_dir(
+        result_directory=result_directory
+    )
+
+
+def read_read_global_sensitivity_analysis_time_set(
```
nicoddemus, line 1548: "read_read"?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1480,64 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    property_id: str, parametric_var_id: str, trend_id: str, results_directory: Path
+) -> Dict[str, numpy.array]:
+    """
+    Read the global sensitivity analysis coefficients results.
+    """
+    uq_meta_data = read_global_sensitivity_analysis_meta_data(results_directory)
+    with open_global_sensitivity_analysis_result_file(results_directory) as result_file:
+        global_sensitivity_analysis_coefficients = result_file[
+            GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+        ]["global_sensitivity_analysis"][:]
```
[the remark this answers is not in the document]
nicoddemus: Didn't mypy catch those? :eyes:

src/alfasim_sdk/result_reader/aggregator.py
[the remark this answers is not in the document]
nicoddemus: @ro-oliveira95 has a point, "this can never occur today because of this code" is not good enough, because "this code" might change in the future, where you assumptions are no longer valid.

Having said that, I think it is fine to just let this blow up with whatever exception will happen when we try to access an invalid group.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1513,72 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    """
+    uq_meta_data = read_global_sensitivity_analysis_meta_data(results_directory)
+    with open_global_sensitivity_analysis_result_file(results_directory) as result_file:
+        if not result_file:
+            return
+        global_sensitivity_analysis_coefficients = result_file[
+            GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+        ]["global_sensitivity_analysis"][:]
+
+    coefficients_key = f"{property_id}::{parametric_var_id}@{trend_id}"
```
ro-oliveira95, line 1576: Probably it would be a good idea to extract this line into a function to be used here and in alfasim, to enforce a pattern.

```python
def obtain_coefficient_key(property_id: str, parametric_var_id: str, trend_id: str) -> str:
    return f"{property_id}::{parametric_var_id}@{trend_id}"
```

prusse-martin (changes requested):

tests/results/test_aggregator.py
```diff
@@ -291,3 +299,109 @@ def test_read_time_sets(
...
+            "temperature::parametric_var_1@trend_id_1"
+        ].parametric_var_name
+        == "A"
+    )
+    assert (
+        global_gsa_meta_data[
+            "temperature::parametric_var_2@trend_id_1"
+        ].parametric_var_name
+        == "B"
+    )
```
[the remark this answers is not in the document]
BeneBr: No, i did not notice that. Thanks bro.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +90,102 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+    :ivar parametric_var_id:
+        The id associated a parametric variable.
+    :ivar parametric_var_name:
+        The name associated a parametric variable.
+    :ivar network_element_name:
+        The name of network where this data is associated.
+    :ivar position:
+        The element position in a pipe.
+    :ivar unit:
+        The unit of property (degC, bar)
```
BeneBr, line 113:
```suggestion
        The unit of property (-, %)
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +90,102 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+class GlobalSensitivityAnalysisMetaInformation:
+    """
+    A class that hold each global sensitivity analysis
+    metadata information.
+    :ivar property:
+        The property name (holdup, total mass flow rate).
+    :ivar trend_id:
+        The id associated a specific trend.
+    :ivar category:
+        The property category (pressure, temperature).
```
BeneBr, line 103:
```suggestion
        The property category for global sensitivity analysis they will always be dimensionless.
```

ro-oliveira95 (approved):
All good. 

But to facilitate the reviewers to know what entered after our review, try to avoid earlier `squash`s as they obfuscate the new changes. Mentioning this because it was hard for me to know if what I pointed out were addressed.

src/alfasim_sdk/result_reader/aggregator.py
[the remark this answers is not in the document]
prusse-martin: The file is put in swmr mode just after all groups and datasets are created, and just after that the file could be read (given the `*.creating` protocol is respected).

src/alfasim_sdk/result_reader/aggregator.py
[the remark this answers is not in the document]
prusse-martin: Technically the key should not be constructed. It should come from the existing metadata. If it is in the metadata the group/dset should exist.

This design is not an implementation detail. Is the protocol.

Those id's have no clear meaning to sdk. It is actually possible to access the metadata and get then, but we usually offer a high level interface to abstract that:

https://github.com/ESSS/alfasim-sdk/blob/master/src/alfasim_sdk/result_reader/reader.py#L147-L170

That `trend_metadata` has those ids. But most ids (property is the exception) are internal alfasim detailts.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1544,70 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+        ]["global_sensitivity_analysis"][:]
+
+    meta_information = uq_meta_data.global_sensitivity_analysis_meta_data[
+        coefficients_key
+    ]
+    return {
+        coefficients_key: global_sensitivity_analysis_coefficients[
+            meta_information.qoi_index, meta_information.qoi_data_index
+        ]
+    }
```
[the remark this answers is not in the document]
BeneBr: Sorry, this point was bypassing for me. Really we dont need the key. Thanks @ro-oliveira95 and @prusse-martin.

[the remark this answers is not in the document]
BeneBr: If i do as you mentioned, i take this error:
```shell
tests/results/test_aggregator.py:326 (test_read_uq_global_sensitivity_analysis)
global_sa_results_dir = PosixPath('/tmp/pytest-of-gustavopb/pytest-91/test_read_uq_global_sensitivit0/test_aggregator/fake_uq_dir')

    def test_read_uq_global_sensitivity_analysis(global_sa_results_dir: Path) -> None:
>       data = read_global_sensitivity_coefficients(coefficients_key="temperature::parametric_var_1@trend_id_1",results_directory=global_sa_results_dir)

test_aggregator.py:328: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../src/alfasim_sdk/result_reader/aggregator.py:1557: in read_global_sensitivity_coefficients
    return global_sensitivity_analysis_coefficients[meta.qoi_index,  meta.qoi_data_index]
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
../../../../Work/miniconda/envs.3/alfasim_gui-py310/lib/python3.10/site-packages/h5py/_hl/dataset.py:766: in __getitem__
    if self._fast_read_ok and (new_dtype is None):
../../../../Work/miniconda/envs.3/alfasim_gui-py310/lib/python3.10/site-packages/h5py/_hl/base.py:534: in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
../../../../Work/miniconda/envs.3/alfasim_gui-py310/lib/python3.10/site-packages/h5py/_hl/dataset.py:745: in _fast_read_ok
    self._extent_type == h5s.SIMPLE
../../../../Work/miniconda/envs.3/alfasim_gui-py310/lib/python3.10/site-packages/h5py/_hl/base.py:534: in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
../../../../Work/miniconda/envs.3/alfasim_gui-py310/lib/python3.10/site-packages/h5py/_hl/dataset.py:634: in _extent_type
    return self.id.get_space().get_simple_extent_type()
h5py/_objects.pyx:54: in h5py._objects.with_phil.wrapper
    ???
h5py/_objects.pyx:55: in h5py._objects.with_phil.wrapper
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   ValueError: Invalid dataset identifier (invalid dataset identifier)
```
Só to access the an specific data i do this: ```result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]['global_sensitivity_analysis'][meta.qoi_index,  meta.qoi_data_index]``` and works good.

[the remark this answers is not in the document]
prusse-martin: ah... That is because I was indexing the dset after the file has been closed the correct think I should suggest was:

```suggestion
    metadata = read_global_sensitivity_analysis_meta_data(results_directory)
    meta = metadata.global_sensitivity_analysis_meta_data[coefficients_key]
    
    with open_global_sensitivity_analysis_result_file(results_directory) as result_file:
        if not result_file:
            return
        coefficients_dset = result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME][
            "global_sensitivity_analysis"
        ]
        return {
            coefficients_key: coefficients_dset[meta.qoi_index, meta.qoi_data_index]
        }

File: src/alfasim_sdk/result_reader/aggregator.py [hunk 24]
```diff
@@ -85,6 +91,112 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+                if not result_file:
+                    return None
+
+                loaded_metadata = json.loads(
+                    result_file[META_GROUP_NAME].attrs["global_sensitivity_analysis"]
+                )
+                return cls(
+                    global_sensitivity_analysis_meta_data=map_data(loaded_metadata),
+                    result_directory=result_directory,
+                )
```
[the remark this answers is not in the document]
BeneBr: Hmmmm, yes you are right. For keep the consistence with previous implemented same cases, 1 and 2 will return an empty metadata.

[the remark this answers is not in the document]
nicoddemus: I agree. Note that in that case you do not need the first `if`, you can let `open_global_sensitivity_analysis_result_file` handle that for you.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1433,3 +1557,72 @@ def merge_metadata(a_metadata, b_metadata, *, source, update=(), min_=(), max_=(
...
+    if not metadata.global_sensitivity_analysis_meta_data:
+        return
+    meta = metadata.global_sensitivity_analysis_meta_data[coefficients_key]
+    with open_global_sensitivity_analysis_result_file(results_directory) as result_file:
+        if not result_file:
+            return
+        coefficients_dset = result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME][
+            "global_sensitivity_analysis"
+        ]
+        return coefficients_dset[meta.qoi_index, meta.qoi_data_index]
```
prusse-martin, line 1628:
```suggestion
def read_global_sensitivity_coefficients(
    coefficients_key: str,
    metadata: GlobalSensitivityAnalysisMetadata,
) -> Optional[np.ndarray]:
    """
    Read the global sensitivity analysis coefficients results.
    """
    # The metadata is empty.
    if not metadata.global_sensitivity_analysis_meta_data:
        return

    meta = metadata.global_sensitivity_analysis_meta_data[coefficients_key]
    with open_global_sensitivity_analysis_result_file(metadata.result_directory) as result_file:
        if not result_file:
            return

        gsa_group = result_file[GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME]
        coefficients_dset = gsa_group["global_sensitivity_analysis"]
        return coefficients_dset[meta.qoi_index, meta.qoi_data_index]
```

prusse-martin (approved):

nicoddemus (approved):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +91,124 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+            qoi_index=data["qoi_index"],
+            qoi_data_index=data["qoi_data_index"],
+        )
+
+    @classmethod
+    def get_metadata_from_dir(
+        cls, result_directory: Path
+    ) -> Optional["GlobalSensitivityAnalysisMetadata"]:
+        """
+        Return the metadata info from result file.
```
nicoddemus, line 187:
```suggestion
        Return the metadata info from result directory. If the directory does not exist/is invalid, return an empty metadata.
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +91,124 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+            position_unit=data["position_unit"],
+            unit=data["unit"],
+            qoi_index=data["qoi_index"],
+            qoi_data_index=data["qoi_data_index"],
+        )
+
+    @classmethod
+    def get_metadata_from_dir(
+        cls, result_directory: Path
+    ) -> Optional["GlobalSensitivityAnalysisMetadata"]:
```
nicoddemus, line 185:
```suggestion
    ) -> "GlobalSensitivityAnalysisMetadata":
```

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +91,124 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+            category=data["category"],
+            parametric_var_id=data["parametric_var_id"],
+            parametric_var_name=data["parametric_var_name"],
+            network_element_name=data["network_element_name"],
+            position=data["position"],
+            position_unit=data["position_unit"],
+            unit=data["unit"],
+            qoi_index=data["qoi_index"],
+            qoi_data_index=data["qoi_data_index"],
+        )
```
nicoddemus, line 180:
```suggestion
    @classmethod
    def from_dict(
        cls, data: dict[str, Any]
    ) -> Self:
        return cls(
            property_id=data["property_id"],
            trend_id=data["trend_id"],
            category=data["category"],
            parametric_var_id=data["parametric_var_id"],
            parametric_var_name=data["parametric_var_name"],
            network_element_name=data["network_element_name"],
            position=data["position"],
            position_unit=data["position_unit"],
            unit=data["unit"],
            qoi_index=data["qoi_index"],
            qoi_data_index=data["qoi_data_index"],
        )
```

prusse-martin: https://docs.python.org/3/library/typing.html#typing.Self

> New in version 3.11.

We can't use that because alfasim-sdk supports python 3.8, 3.9, and 3.10, and I don't think there is a future for `Self`.

nicoddemus: It is importable from `typing_extensions` and supported by all those version, please let's use it.

BeneBr: The ```from_dict``` method iis defined in ```GlobalSensitivityAnalysisMetadata``` but they return another type od data: ```GlobalSensitivityAnalysisMetaInformation```. So it could not be used.

src/alfasim_sdk/result_reader/aggregator.py
BeneBr: And as @prusse-martin has been pointed, the ```Self``` is not supoorted. I tryied to do this and this error occur:
```shell
ImportError while loading conftest '/home/users/gustavopb/Projects/alfasim-sdk/tests/conftest.py'.
../conftest.py:12: in <module>
    from alfasim_sdk.result_reader.aggregator_constants import META_GROUP_NAME, GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME, \
../../src/alfasim_sdk/result_reader/__init__.py:1: in <module>
    from .aggregator import ALFASimResultMetadata
../../src/alfasim_sdk/result_reader/aggregator.py:132: in <module>
    class GlobalSensitivityAnalysisMetadata:
../../src/alfasim_sdk/result_reader/aggregator.py:161: in GlobalSensitivityAnalysisMetadata
    def get_metadata_from_dir(cls, result_directory: Path) -> self:
E   NameError: name 'self' is not defined
```

src/alfasim_sdk/result_reader/aggregator.py
nicoddemus: Ahh definitely, my bad.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +92,124 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+
+@attr.s(slots=True, hash=False)
+class GlobalSensitivityAnalysisMetadata:
+    """
+    A class that hold the global sensitivity analysis metadata.
+    """
+
+    global_sensitivity_analysis_meta_data: Dict[
+        str, GlobalSensitivityAnalysisMetaInformation
+    ] = attr.ib(validator=attr.validators.instance_of(Dict))
```
nicoddemus, line 172: This is a bit confusing:

1. `GlobalSensitivityAnalysisMetaInformation` and `GlobalSensitivityAnalysisMetadata` are nearly synonymous, so it is hard to tell them apart just from the name, without looking at the class definition/docs.
2. `GlobalSensitivityAnalysisMetadata.global_sensitivity_analysis_meta_data` is a dict of `GlobalSensitivityAnalysisMetaInformation`... this should send warning signals to you, having a class variable which is the name of the class itself, but references actually another data type.

I think `GlobalSensitivityAnalysisMetadata` is fine, but we should rename `GlobalSensitivityAnalysisMetaInformation` to something else. Doesn't it hold information about one of the parametric variables in a Global SA? How about `ParametricVariableMetadata`? If so, rename the `global_sensitivity_analysis_meta_data` as well to reflect that, `parametric_variables_metadata`.

nicoddemus (changes requested):

src/alfasim_sdk/result_reader/aggregator.py
BeneBr: I renamed it to ```QuantityOfInterestMetaData``` that what it really represent.

nicoddemus: Sounds good!

BeneBr: Humm, i agree.

nicoddemus: No I was not aware, I agree an intermediate "item" container would be good.

nicoddemus: Actually I'm not sure I follow, but I will defer to your decision here.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -85,6 +92,120 @@ class ResultsNeedFullReloadError(RuntimeError):
...
 
 
+@attr.s(slots=True, hash=False)
+class GlobalSensitivityAnalysisMetadata:
+    """
+    A class that hold the global sensitivity analysis metadata.
+    """
+
+    @attr.s(slots=True, hash=False)
+    class GSAItem:
```
nicoddemus, line 102: @prusse-martin please double check if this is what you meant by your previous comment. 👍 

https://github.com/ESSS/alfasim-sdk/pull/357#discussion_r1482897930

nicoddemus (approved):

src/alfasim_sdk/result_reader/aggregator.py
prusse-martin: I was more about a module level `GlobalSensitivityAnalysisMetadataItem` or a class scope (like it is here) `GlobalSensitivityAnalysisMetadata.Item`.

But `GlobalSensitivityAnalysisMetadata.GSAItem` is OK.

PR #358, opened 2024-01-26
Add invoke tasks to a collection

ThalesCarl:
Previously, the invoke tasks were defined in the global namespace. This PR aims to add them to a collection in order to facilitate the overloading of tasks by the user plugins and to better organize them. 

Off-topic: I've also included in this PR a check if the `update` function successfully created the `hook_specs.h` file.

ASIM-5375


src/alfasim_sdk/default_tasks.py
```diff
@@ -3,16 +3,49 @@
...
 from colorama import Fore
 from colorama import Style
 from hookman.hookman_generator import HookManGenerator
+from invoke import Collection
 from invoke import Exit
+from invoke import Task
 from invoke import task
 from strictyaml.ruamel import YAML
 
+sdk = Collection()
```
ThalesCarl, line 19: I've called the collections `sdk` instead of `sdk_tasks`, because when we call then we will use `sdk.tasks['example']` instead of `sdk_tasks.tasks['example']`, which seemed a little bit of repetition.

What do you guys think?

nicoddemus: Not sure, I find the symbol `sdk`, in a project called `alfasim-sdk`, to be a *collection of sdk tasks* to be a bit surprising... 

But OK, not sure if this will be a problem either, go for it.

nicoddemus (approved):

src/alfasim_sdk/default_tasks.py
ThalesCarl: I agree, but I cannot think of a better name. Perhabs `sdk_collection` or `sdk_namespace` (`sdk_ns` for short) but it felt strange as well.

src/alfasim_sdk/default_tasks.py
```diff
@@ -256,9 +289,13 @@ def update(ctx):
     plugin_id = plugin_folder.name
     plugin_location = plugin_folder.parent
     hm.generate_hook_specs_header(plugin_id, plugin_location)
+    assert hook_specs_file_path.is_file()
+    print_message(
+        "Successfully updated alfasim-sdk's files", color=Fore.GREEN, bright=True
+    )
```
prusse-martin, line 295: What if the update failed but an old header is in place? (I am not sure that can happen)

Instead of an `assert` and if to print a better message would be nice:

```suggestion
    if hook_specs_file_path.is_file():
        print_message(
            "Successfully updated alfasim-sdk's files", color=Fore.GREEN, bright=True
        )
    else:
        print_message(
            "Failed to update alfasim-sdk's files", color=Fore.RED, bright=True
        )
        raise invoke.Exit(message=None, code=1)  # `code != 0`
```
(test for stale file not present)

prusse-martin (approved):

src/alfasim_sdk/default_tasks.py
nicoddemus: `sdk_ns` I like

ggrbill (approved):

src/alfasim_sdk/default_tasks.py
ThalesCarl: That's true. I've added a snippet for removing the older file, if present, before calling `hm.generate_hook_specs_header`.

Also, the check I was doing `hook_specs_file_path.is_file()` is not valid because this variable's value is actually `alfasim-sdk/_internal/hook_specs.py` used to generate the `hook_specs.h` file, instead of the file itself. I've added a new variable `generated_hook_file` that is the file we need to check.

prusse-martin: You could add a `# pragma: no cover`, or do some mocking:

```python
def test_eerro_message(capfd, mocker):
    mocker.patch.object(HookManGenerator, 'generate_hook_specs_header', autospec=True)
    with pytest.raises(invoke.Exit) as exit_error:
        update(mocker.MagicMock())
    assert exit_error.code != 0
    stdout, stderr = capfd.readouterr()
    assert stdout == 'the message'
    assert stderr == ''
```

ThalesCarl: I tried mocking, but the mock does not work because we need to call a subprocess for running invoke. Furthermore, I'm not able to call the `update` function directly, because this function is a `invoke` `Task` that needs a `ctx` argument, which is not available.

Hence, I've decided to go for using the `#pragma: no cover`. Even tough, I'm still not sure if the `else` block will be reachable in any circumstance, because the only place an error might occur is at the call of `hm.generate_hook_specs_header`, which will cause the code to return before reaching that block.

alexandrebbruno (commented):
great job

PR #361, opened 2024-02-22
Add History Matching metadata and support for result file opening

ro-oliveira95:
ASIM-5599


ro-oliveira95:
Closing as these changes actually belong to the "read history matching data" context and will enter next, along with other things.

PR #364, opened 2024-03-01
[ASIM-5600] Read History Matching results

ro-oliveira95:
Add History Matching metadata and reading functions to aggregator to allow reading HM results data.

ASIM-5600

### DIFF
--- a/src/alfasim_sdk/result_reader/aggregator.py
+++ b/src/alfasim_sdk/result_reader/aggregator.py
@@ -1722,7 +1722,7 @@ def read_history_matching_metadata(result_directory: Path) -> HistoryMatchingMet
 def read_history_matching_result(
     metadata: HistoryMatchingMetadata,
     hm_type: Literal["deterministic", "probabilistic"],
-    hm_result_key: HistoryMatchingResultKeyType | None = None,
+    hm_result_key: Optional[HistoryMatchingResultKeyType] = None,
 ) -> Dict[HistoryMatchingResultKeyType, np.ndarray | float]:
     """
     :param metadata:
--- a/src/alfasim_sdk/result_reader/aggregator.py
+++ b/src/alfasim_sdk/result_reader/aggregator.py
@@ -10,6 +10,7 @@ from typing import DefaultDict
 from typing import Dict
 from typing import Iterator
 from typing import List
+from typing import Literal
 from typing import Optional
 from typing import Tuple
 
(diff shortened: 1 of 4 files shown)


prusse-martin (approved):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1615,9 +1703,87 @@ def read_global_sensitivity_coefficients(
...
+    ignored_file = result_directory / f"{result_filename}.creating"
+
+    if not filepath.is_file():
+        yield None
+    # Avoid to read result files with incomplete metadata.
+    elif not ignored_file.is_file():
+        with _open_result_file(filepath) as file:
+            yield file
+    else:
+        yield None
```
nicoddemus, line 1789: Minor, but I would keep each condition in their own `if`:

```suggestion
    filepath = result_directory / result_filename
    creating_file = result_directory / f"{result_filename}.creating"

    if not filepath.is_file():
        yield None
    # Do not open incomplete files.
    if creating_file.is_file():
        yield None
    else:
        with _open_result_file(filepath) as file:
            yield file
```

The way it is written is a bit more convoluted than it needs to be.

tests/conftest.py
```diff
@@ -259,3 +269,100 @@ def global_sa_results_dir(datadir: Path) -> Path:
...
+            "objective_functions": objective_functions,
+            "data_index": 1,
+        },
+    }
+
+    meta_group.attrs[HISTORY_MATCHING_GROUP_NAME] = json.dumps(fake_meta)
+    dataset[:] = result
+
+    file.swmr_mode = True
+    file.close()
```
nicoddemus, line 322: Always mind explicit "close" calls like this: this will never be reached if an exception happens until we get to this point.

You should use a context manager or try/finally to ensure the file gets closed even if an exception happens.

tests/conftest.py
```diff
@@ -259,3 +269,100 @@ def global_sa_results_dir(datadir: Path) -> Path:
...
+        limits=limits,
+    )
+
+    return result_dir
+
+
+@pytest.fixture()
+def hm_deterministic_results_dir(datadir: Path) -> Path:
+    """
+    Crete a History Matching result folder with a populated result file for each type of analysis
```
nicoddemus, line 352:
```suggestion
    Create a History Matching result folder with a populated result file for each type of analysis
```

But the doc in general is incorrect, it is only creating one type of analysis here.

tests/conftest.py
```diff
@@ -259,3 +269,100 @@ def global_sa_results_dir(datadir: Path) -> Path:
...
+    dataset[:] = result
+
+    file.swmr_mode = True
+    file.close()
+
+
+@pytest.fixture()
+def hm_probabilistic_results_dir(datadir: Path) -> Path:
+    """
+    Crete a History Matching result folder with a populated result file for each type of analysis
```
nicoddemus, line 328:
```suggestion
    Create a History Matching result folder with a populated result file for each type of analysis
```

But the doc in general is incorrect, it is only creating one type of analysis here.

nicoddemus (approved):

tests/conftest.py
ro-oliveira95: Good catch.

tests/conftest.py
ro-oliveira95: You're correct. I missed this one while refactoring things.

ro-oliveira95:
Ops, I accidentally merged this PR without addressing the review appointments.
Will address then in the base branch's PR https://github.com/ESSS/alfasim-sdk/pull/369.

nicoddemus:
> Will address then in the base branch's PR https://github.com/ESSS/alfasim-sdk/pull/369.

Not sure that's what you meant, but please create a new PR targeting #369.

PR #366, opened 2024-03-06
Add a HOOK capable of updating plugin variables only once at the end of the time step computation

alexandrebbruno:
[DCC-208](https://esss.atlassian.net/browse/DCC-208l)

Add a HOOK capable of updating plugin variables only once at the end of the time step computation (not on each step of Newton method - which is done by regular HOOK_UPDATE_PLUGINS_SECONDARY_VARIABLES)

[DCC-208]: https://esss.atlassian.net/browse/DCC-208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/alfasim_sdk/_internal/hook_specs.py
+++ b/src/alfasim_sdk/_internal/hook_specs.py
@@ -142,7 +142,7 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
             int errcode = -1;
             int size_U = -1;
             int size_E = -1;
-            int liq_id = -1;
+            int oil_id = -1;
             errcode = alfasim_sdk_api.get_field_id(
                 ctx, &oil_id, "oil");
             double* vel;
@@ -152,7 +152,7 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
                 TimestepScope::CURRENT
             }
             errcode = alfasim_sdk_api.get_simulation_array(
-                ctx, &vel, (char*) "U", Fields_OnFaces, liq_id, &size_U);
+                ctx, &vel, (char*) "U", Fields_OnFaces, oil_id, &size_U);
             double* kinetic_energy;
             char* name = "kinetic_energy_of_oil";
             int global_idx = 0;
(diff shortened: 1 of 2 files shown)


src/alfasim_sdk/_internal/hook_specs.py
```diff
@@ -178,6 +178,64 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
     """
 
 
+def update_plugins_secondary_variables_time_explicit(ctx: "void*") -> "int":
```
ThalesCarl, line 181: Why are we calling this hook `_time_explicit`?

src/alfasim_sdk/_internal/hook_specs.py
```diff
@@ -178,6 +178,64 @@ def update_plugins_secondary_variables(ctx: "void*") -> "int":
...
+            for (int i =0; i < size_U; ++i){
+                kinetic_energy[i] = vel[i] * vel[i] / 2.;
+            }
+            return OK;
+        }
+
+    In the example above the variable ``kinetic_energy_of_oil`` was registered as a global variable, but its value is
+    obtained for `oil field` and only at the end of the time steps' linear solver (explicit approach). If this
+    variable would be calculated to all fields then the ``global_idx`` would be substituted by ``field_idx`` and it
+    would be performed to each `field`.
```
ThalesCarl, line 235: How does the value is obtained for the oil_field if the variable is global and the `oil_id` variable is not used?

ThalesCarl (approved):
Great work. I just have some conceptual questions.

src/alfasim_sdk/_internal/hook_specs.py
alexandrebbruno: Putz
I made this example based on the one used at `update_plugins_secondary_variables`, but I think that line [221](https://github.com/ESSS/alfasim-sdk/pull/366/files/990d9b4b09f0cd1088f46847624c7142d02f3a36#diff-57379b1558ae80d93105233310befbcc92749b75f1add1edbe7c699a0b2b3700R211) should be using `oil_id` instead of `liq_id` and the text is saying that we are only computing the kinetic_energy of oil field because we are using only de oil velocity 🤔

ThalesCarl: Nice, thanks for correcting it.

ggrbill (approved):
Good Job!

caio-ramos (approved):

eniohayashi (approved):

PR #371, opened 2024-03-14
Expose volumetric flow rate that can be used by plugins

alexandrebbruno:
Update documentation that expose volumetric flow rate that can be used by plugins

https://esss.atlassian.net/browse/DCC-195


ggrbill (approved):
Nice!

PR #373, opened 2024-04-02
Add Alfasim version info to CaseDescription

cauebs:


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -3407,13 +3408,23 @@ class NumericalOptionsDescription:
     enable_fast_compositional: bool = attr.ib(default=True, validator=instance_of(bool))
 
 
+@attr.s(frozen=True)
+class AlfasimVersionInfo:
+    platform: str = attr.ib()
+    version: str = attr.ib()
+    revision: str = attr.ib()
+    date: str = attr.ib()
```
nicoddemus, line 3416:
```suggestion
class AlfasimVersionInfo:
    """
    Information about the ALFAsim release used to generate the alfacase file.
    
    Used for informational purposes only.
    """
    
    #: Platform alfasim was running. 
    platform: str = attr.ib()
    
    #: Release version.
    version: str = attr.ib()
    
    #: Internal revision hash.    
    revision: str = attr.ib()
    
    #: Release date in 'YYYY-MM-DD h:m:s' format.
    date: str = attr.ib()
```

nicoddemus (approved):
Nice work!

ro-oliveira95 (approved):

src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
```diff
@@ -341,6 +341,19 @@ def attrib_instance(type_: type) -> attr._make._CountingAttr:
...
+    """
+    Create a new optional attr attribute with validator for the given type_.
+    """
+    metadata = {"type": "instance", "class_": Optional[type_]}
+    return attr.ib(
+        default=None,
+        validator=optional(instance_of(type_)),
+        type=Optional[type_],
+        metadata=metadata,
+    )
```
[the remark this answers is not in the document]
prusse-martin: Also.
For example `attrib_curve` we got an `is_optional` flag instead of having an `attrib_optional_curve`:

https://github.com/ESSS/alfasim-sdk/blob/08b36bd85eea6a93187fed0b914dcb7f0abd4498/src/alfasim_sdk/_internal/alfacase/case_description_attributes.py#L273-L328

Maybe adding that flag to `attrib_instance` will make maintenance easier if we need to change the metadata contents.

docs/source/alfacase/api_reference.rst
```diff
@@ -12,6 +12,8 @@ Case
 
 .. autoclass:: alfasim_sdk.CaseDescription()
 
+.. autoclass:: alfasim_sdk.AlfasimVersionInfo()
```
prusse-martin, line 15: https://alfasim-sdk--373.org.readthedocs.build/en/373/alfacase/api_reference.html#alfasim_sdk.CaseDescription

image

The generated docs is missing a link for `AlfasimVersionInfo` in `CaseDescription` definition =(
At first look I didn't notice anything missing. Could you recheck that?

prusse-martin (changes requested):

BeneBr (approved):

docs/source/alfacase_definitions/CaseDescription.txt
```diff
@@ -6,6 +6,8 @@
 
         class CaseDescription
             name: \ :class:`Optional <typing.Optional>`\[str] = None
+            alfasim_version_info: \ :class:`Optional <typing.Optional>`\[AlfasimVersionInfo] = None
```
prusse-martin, line 9: Maybe try
```suggestion
            alfasim_version_info: \ :class:`Optional <typing.Optional>`\[\ :class:`AlfasimVersionInfo <AlfasimVersionInfo>`\] = None
```

cauebs: Problem is that this file is generated. If I manually edit it, the build will fail.

prusse-martin (approved):
I will approve, but plz try this before merging.

docs/source/alfacase_definitions/CaseDescription.txt
prusse-martin: > Problem is that this file is generated. If I manually edit it, the build will fail.

Given this is generated to proper way would be fixing the generator.

cauebs: > Given this is generated to proper way would be fixing the generator.

Of course.

I tried what you suggested. You were right as to what was missing, and this is what it ended up like:
```python
elif is_attrs(ref_value):
    name = attrs_formatted(ref_value)
```

PR #388, opened 2024-05-28
Bump version to 2024.2

prusse-martin:


cauebs (approved):

src/alfasim_sdk/_internal/version.py
```diff
@@ -1 +1 @@
-__version__ = "2024.1"
+__version__ = "2024.2"
```
nicoddemus, line 1: How about using a `.dev` suffix to signal this has not been released yet?

```suggestion
__version__ = "2024.2.dev"
```

nicoddemus (approved):

ro-oliveira95 (approved):

src/alfasim_sdk/_internal/version.py
prusse-martin: OK... I will update the alfasim release process to mention we should update the sdk version when starting an alfasim RB.

prusse-martin:
@nicoddemus 
From alfasim build after the introduction of `.dev` in version:
```
E               AssertionError: Error messages not found in expected:
E                   "Unsupported version of "hydrates" plugin, current ALFAsim version uses alfasim-sdk="2024.2.dev0" and plugin has the following requirement=">=2024.2""
```
The version spec used when creating the packages is `">=MAJOR.MINOR"` and `MAJOR.MINOR.devX < MAJOR.MINOR` I will add the full version on packages from now on.
I could drop the `.devX` part when comparing in alfasim but this looks safer.

nicoddemus:
> The version spec used when creating the packages is ">=MAJOR.MINOR" and MAJOR.MINOR.devX < MAJOR.MINOR I will add the full version on packages from now on.

Why is a plugin depending on `>=2024.2` which has not even been released yet? Shouldn't it depend on `>=2024.1`, which is the latest release?

prusse-martin:
> Why is a plugin depending on >=2024.2 which has not even been released yet? Shouldn't it depend on >=2024.1, which is the latest release?

Because we did just release `2024.1`, plugins build with that version will depend on `2024.1`, sdk just did enter the development phase of `2024.2`.

BeneBr (approved):

PR #390, opened 2024-06-14
Update the sdk bootstrap code to fail fast on clear problems with the lib from alfasim

prusse-martin:
The problems covered in this update:

- Can not load the lib (not found or corrupted);
- Can not fetch the address of a function from the lib;

ASIM-5532


src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_linux.h
```diff
@@ -44,56 +44,80 @@ inline int alfasim_sdk_open(ALFAsimSDK_API* api)
...
+        dlerror();\
+        api->func_name = (func_type)dlsym(api->handle, #func_name);\
+        char *last_error = dlerror();
+        if (last_error != NULL) {\
+            dlclose(api->handle);\
+            api->handle = nullptr;\
+            return SDK_FAILED_TO_LOAD_FUNCTION;\
+        }\
+    }
+    #define LOAD_SDK_PROC(func_name) LOAD_SDK_PROC_MAP_TYPE(func_name, func_name ## _func)
```
prusse-martin, line 72: It is many years since I wrote any macro.
Is this (and the ones in the other bootstrap file) OK?

eniohayashi (approved):
Nice!!

src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_linux.h
```diff
@@ -44,56 +44,80 @@ inline int alfasim_sdk_open(ALFAsimSDK_API* api)
...
+    // > a NULL return from dlsym() need not indicate an error.  The correct way to distinguish an
+    // > error from a symbol whose value is NULL is to call dlerror(3) to clear any old error
+    // > conditions, then call dlsym(), and then call dlerror(3) again, saving its return value
+    // > into a variable, and check whether this saved value is not NULL.
+    #define LOAD_SDK_PROC_MAP_TYPE(func_name, func_type) {\
+        dlerror();\
+        api->func_name = (func_type)dlsym(api->handle, #func_name);\
+        char *last_error = dlerror();\
+        if (last_error != NULL) {\
+            dlclose(api->handle);\
```
nicoddemus, line 67: This should be called even if an error happened? Won't `api->handle` be nullptr or point to invalid memory in that case?

src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_win.h
```diff
@@ -68,56 +68,74 @@ inline int alfasim_sdk_open(ALFAsimSDK_API* api)
...
+    if (api->handle == NULL) {
+        return SDK_FAILED_TO_LOAD_DLL;
+    }
+
+    // https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getprocaddress
+    // > If the function fails, the return value is NULL.
+    #define LOAD_SDK_PROC_MAP_TYPE(func_name, func_type) {\
+        api->func_name = (func_type)GetProcAddress(api->handle, #func_name);\
+        if (api->func_name == NULL) {\
+            FreeLibrary(api->handle);\
```
nicoddemus, line 85: Same comment here.

nicoddemus (approved):
Nice work!

src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_linux.h
prusse-martin: At this point the handle is valid (`dlopen` was OK), `dlsym` just failed to load a symbol from the library. This `dlclose` free the resources from that successful `dlopen`.

src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_win.h
prusse-martin: "Same" (windows not linux, but the logic is the same).

src/alfasim_sdk/alfasim_sdk_api/detail/bootstrap_linux.h
nicoddemus: ahh my bad, thanks.

prusse-martin:
IIRC the "cpp wrapper" bootstrap the sdk in static initialization and the build and packing tools from sdk will try to load the plugin dll in a context where the api can't be bootstrapped causing it to fail.

But without this the plugin execution could error out at some point while alfasim is executing the simulation (say with a seemingly random access violation) because bootstrap would fail silently.

nicoddemus:
Can't you work to get this merged?

eniohayashi:
> @eniohayashi @ThalesCarl

I think it could be merged. @prusse-martin Was there any problem that made you postpone it?

prusse-martin:
> > @eniohayashi @ThalesCarl
> 
> I think it could be merged. @prusse-martin Was there any problem that made you postpone it?

Yes.
> IIRC the "cpp wrapper" bootstrap the sdk api in static initialization and the build and packing tools from sdk will try to load the plugin dll in a context where the api can't be bootstrapped causing it to fail.

https://github.com/ESSS/alfasim-sdk/pull/390#issuecomment-2636860304

PR #393, opened 2024-07-05
[ASIM-5674] - Read uncertainty propagation results

BeneBr:
ASIM-5674


src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -179,13 +269,13 @@ def from_dict(cls, data: Dict[str, Any]) -> Self:
                 qoi_data_index=data["qoi_data_index"],
             )
 
-    gsa_items: Dict[str, GSAItem] = attr.ib(validator=attr.validators.instance_of(Dict))
+    items: Dict[str, GSAItem] = attr.ib(validator=attr.validators.instance_of(Dict))
```
nicoddemus, line 272: :+1: 

We are already on the "Global SA" context because of the class, no need for the attribute to carry that `gsa` prefix.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -205,18 +295,52 @@ def map_data(
...
+    Holder for each uncertainty propagation result.
+    """
+
+    category: str = attr.ib(
+        validator=attr.validators.optional(attr.validators.instance_of(str))
+    )
+    unit: str = attr.ib(
+        validator=attr.validators.optional(attr.validators.instance_of(str))
+    )
+    realization_output: List = attr.ib(default=attr.Factory(List))
```
nicoddemus, line 323: List of what? Can we have a more narrow type here?

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -205,18 +295,52 @@ def map_data(
...
+    std_result: np.ndarray = attr.ib(default=attr.Factory(lambda: np.array([])))
+    mean_result: np.ndarray = attr.ib(default=attr.Factory(lambda: np.array([])))
+
+
+def read_results_file(
+    result_directory: Path,
+    metadata_class: Type[UQMetadataClass],
+    map_data: Callable,
+    meta_data_attrs: str,
+) -> UQMetadataClass:
```
nicoddemus, line 333: See if this works:

```suggestion
MetadataClassType = TypeVar("MetadataClassType", bound=UQMetadataClass)

def read_results_file(
    result_directory: Path,
    metadata_class: Type[MetadataClassType],
    map_data: Callable,
    meta_data_attrs: str,
) -> MetadataClassType:
```

This will provide the correct type narrowing.

nicoddemus (approved):

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -105,57 +121,131 @@ class ResultsNeedFullReloadError(RuntimeError):
     """
 
 
+@attr.s(slots=True, hash=False)
+class BaseUQMetaData:
+    """
+    Base class for UQ metadata analysis.
+    :ivar property_id:
+        The property name (holdup, total mass flow rate).
```
prusse-martin, line 129:
```suggestion
    """
    Base class for UQ metadata analysis.

    :ivar property_id:
        The property name (holdup, total mass flow rate).
```
This is not the first reminder about that empty line in the docs to isolate the sort descripton from the body of the documentation.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -105,57 +121,131 @@ class ResultsNeedFullReloadError(RuntimeError):
...
+class UncertaintyPropagationAnalysesMetaData:
+    """
+    A class that hold the uncertainty propagation analyses metadata.
+    """
+
+    @attr.s(slots=True, hash=False)
+    class UPItem(BaseUQMetaData):
+        """
+        A class that hold each uncertainty propagation analyses metadata.
+        :ivar samples:
```
prusse-martin, line 170:
```suggestion
    class UPItem(BaseUQMetaData):
        """
        A class that hold each uncertainty propagation analyses metadata.

        :ivar samples:
```

prusse-martin (approved):

src/alfasim_sdk/result_reader/aggregator.py
BeneBr: I know that. I'm paying attention on this, but sometimes, some cases passes.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -1692,16 +1822,73 @@ def read_global_sensitivity_analysis_meta_data(
...
+            mean_result=mean_result,
+            std_result=std_result,
+            category=meta.category,
+            unit=meta.unit,
+        )
+
+
+def read_uq_time_set(result_directory: Path, group_name: str) -> Optional[numpy.array]:
+    """
+    Get the time set for based uq analysis results.
```
ro-oliveira95, line 1880: This doesn't apply to all UQ analysis, so my recommendation is to make it clear in the docstrings. Same for the `BaseUQMetadata` class.

```suggestion
def read_uq_time_set(result_directory: Path, group_name: str) -> Optional[numpy.array]:
    """
    Get the time set for uq-based analysis results (Global Sensitivity Analysis or Uncertainty Propagation).
```

tests/conftest.py
```diff
@@ -269,6 +284,99 @@ def global_sa_results_dir(datadir: Path) -> Path:
...
+                    i + (qoi_index * 10) + tl
+                    for tl in (0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7)
+                ]
+                for qoi_index, trend_property in enumerate(
+                    [("trend_id_1", "temperature"), ("trend_id_1", "absolute_pressure")]
+                )
+            ]
+            for i in range(5)
+        ]
+    )
```
ro-oliveira95, line 312: This is probably a case where nested `for`s would be better to read/understand than list comprehensions. Also I recommend replacing vars `i` and `tl` by real words here to make it clear what they represents.

tests/results/test_aggregator.py
```diff
@@ -503,3 +514,84 @@ def test_read_history_matching_result_data(
...
+    )
+    assert np.all(
+        result.mean_result
+        == np.array([12.1, 12.2, 12.3, 12.4, 12.5, 12.6, 12.7], dtype=np.float32)
+    )
+    assert np.allclose(
+        result.std_result,
+        np.array([1.414, 1.414, 1.414, 1.414, 1.414, 1.414, 1.414], dtype=np.float32),
+        rtol=0.001,
+    )
```
ro-oliveira95, line 597: Maybe you could use a data regression check here.

ro-oliveira95 (approved):

tests/conftest.py
nicoddemus: I disagree, particularly I find that list comprehension quite readable.

About the variable names I think it is up to the caller, but in list comprehensions I'm at least used to short named variables: given list comprehensions have a very narrow scope, single letter-vars in there are considered OK.

tests/results/test_aggregator.py
nicoddemus: Best use `num_regression`, which has a tolerance to compare floats.

BeneBr: As you wish.

PR #398, opened 2024-08-21
Add a possibility to plugin user gets wall properties

alexandrebbruno:
PR description:
- Adds get_wall_properties, the method that a plugin developer user will use to get wall properties.

### DIFF
--- a/src/alfasim_sdk/alfasim_sdk_api/api.h
+++ b/src/alfasim_sdk/alfasim_sdk_api/api.h
@@ -865,13 +865,17 @@ DLL_EXPORT int get_wall_properties(void* ctx, double** prop_values, const char* 
     ~~~~~{.cpp}
 
             |
+            |
     [control_volume_1]->[material_name_0; material_name_1; material_name_2]
+            |
             |
     [control_volume_2]->[material_name_0; material_name_1]
             |
+            |
     [control_volume_3]->[material_name_0; material_name_1; material_name_2; material_name_3]
             |
-           \ /(pipe)
+            |
+         (pipe)
 
     errcode = get_wall_material_names(
         ctx, &material_names_in_wall, control_volume_id, &size_wall);
@@ -894,13 +898,17 @@ DLL_EXPORT int get_wall_material_names(void* ctx, char*** material_names_in_wall
     Example of usage:
 
             |
+            |
     [control_volume_1]->[is_fluid_check_0; is_fluid_check_1; is_fluid_check_2]
+            |
             |
     [control_volume_2]->[is_fluid_check_0; is_fluid_check_1]
             |
+            |
     [control_volume_3]->[is_fluid_check_0; is_fluid_check_1; is_fluid_check_2; is_fluid_check_3]
             |
-           \ /(pipe)
+            |
+         (pipe)
 
     errcode = get_wall_material_type(
         ctx, &wall_material, control_volume_id, &size_wall);
--- a/src/alfasim_sdk/alfasim_sdk_api/api.h
+++ b/src/alfasim_sdk/alfasim_sdk_api/api.h
@@ -834,23 +834,17 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
     Example of usage:
 
     ~~~~~{.cpp}
-      [teste]
-            |
-    --[control_volume_1]--[control_volume_2]--[control_volume_3]--> (pipe)
 
-            |
-            |
-    [control_volume_1]->[prop_wall_0; prop_wall_1; prop_wall_2]
-            |
-            |
-    [control_volume_2]->[prop_wall_0; prop_wall_1]
-            |
-            |
-    [control_volume_3]->[prop_wall_0; prop_wall_1; prop_wall_2; prop_wall_3]
-            |
-            |
-         (pipe)
+                                                [prop_wall_4]
+        [prop_wall_3]                           [prop_wall_3]
+        [prop_wall_2]       [prop_wall_2]       [prop_wall_2]
+        [prop_wall_1]       [prop_wall_1]       [prop_wall_1]
+        [prop_wall_0]       [prop_wall_0]       [prop_wall_0]
+              |                   |                   |
+    --[control_volume_1]--[control_volume_2]--[control_volume_3]-->  (Pipe)
+    ~~~~~
 
+    ~~~~~{.cpp}
     errcode = get_wall_properties(
         ctx, &prop_values, "layer_thickness", control_volume_id, &size_wall);
     ~~~~~
@@ -873,19 +867,16 @@ DLL_EXPORT int get_wall_properties(void* ctx, double** prop_values, const char* 
 
     ~~~~~{.cpp}
 
-            |
-            |
-    [control_volume_1]->[material_name_0; material_name_1; material_name_2]
-            |
-            |
-    [control_volume_2]->[material_name_0; material_name_1]
-            |
-            |
-    [control_volume_3]->[material_name_0; material_name_1; material_name_2; material_name_3]
-            |
-            |
-         (pipe)
+                                               [material_name_4]
+      [material_name_3]                        [material_name_3]
+      [material_name_2]   [material_name_2]    [material_name_2]
+      [material_name_1]   [material_name_1]    [material_name_1]
+      [material_name_0]   [material_name_0]    [material_name_0]
+              |                   |                   |
+    --[control_volume_1]--[control_volume_2]--[control_volume_3]-->  (Pipe)
+    ~~~~~
 
+    ~~~~~{.cpp}
     errcode = get_wall_material_names(
         ctx, &material_names_in_wall, control_volume_id, &size_wall);
     ~~~~~
@@ -906,19 +897,16 @@ DLL_EXPORT int get_wall_material_names(void* ctx, char*** material_names_in_wall
 
     Example of usage:
 
-            |
-            |
-    [control_volume_1]->[is_fluid_check_0; is_fluid_check_1; is_fluid_check_2]
-            |
-            |
-    [control_volume_2]->[is_fluid_check_0; is_fluid_check_1]
-            |
-            |
-    [control_volume_3]->[is_fluid_check_0; is_fluid_check_1; is_fluid_check_2; is_fluid_check_3]
-            |
-            |
-         (pipe)
+                                              [is_fluid_check_4]
+      [is_fluid_check_3]                      [is_fluid_check_3]
+      [is_fluid_check_2]  [is_fluid_check_2]  [is_fluid_check_2]
+      [is_fluid_check_1]  [is_fluid_check_1]  [is_fluid_check_1]
+      [is_fluid_check_0]  [is_fluid_check_0]  [is_fluid_check_0]
+              |                   |                   |
+    --[control_volume_1]--[control_volume_2]--[control_volume_3]-->  (Pipe)
+    ~~~~~
 
+    ~~~~~{.cpp}
     errcode = get_wall_material_type(
         ctx, &wall_material, control_volume_id, &size_wall);
     ~~~~~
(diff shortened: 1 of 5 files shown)


ThalesCarl (approved):

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,51 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] wall_names Wall names values array.
+    @param[out] prop_values Wall properties values array.
+    @param[in] prop_name String with the property name. See the list of possible values above.
+    @param[in] control_volume Control volume id.
+    @param[out] size Size of the `wall_names` and `prop_values` array of values.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_wall_properties(void* ctx, char*** wall_names, void** prop_values, const char* prop_name, int control_volume, int* size);
+
```
caio-ramos, line 859: Nice documentation 👍

caio-ramos (approved):

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,51 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] wall_names Wall names values array.
+    @param[out] prop_values Wall properties values array.
+    @param[in] prop_name String with the property name. See the list of possible values above.
+    @param[in] control_volume Control volume id.
+    @param[out] size Size of the `wall_names` and `prop_values` array of values.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_wall_properties(void* ctx, char*** wall_names, void** prop_values, const char* prop_name, int control_volume, int* size);
```
prusse-martin, line 858: DLL_EXPORT int get_wall_properties(void* ctx, char*** wall_names, void** prop_values, const char* prop_name, int control_volume, int* size);

I am against the api returning pre allocated buffers (the caller should allocate those).
But someone put those kind of stuff in the api already... and someone could argue that praticatily is best (I disagree here).

But having an "out" argument as `void *` is just not OK.
`ctx` is a `void *` because the caller is not supposed to type cast it (alfasim supply that as an opaque pointer).
I will not bend on this just because there are examples of that in the api. I don't think those should be there.

Create variations with the correct explicit type, and if called to obtain a property in the wront type return a failure error core.
```c
DLL_EXPORT int get_wall_properties_bool(void* ctx, char*** wall_names, bool** prop_values, const char* prop_name, int control_volume, int* size);
DLL_EXPORT int get_wall_properties_double(void* ctx, char*** wall_names, double** prop_values, const char* prop_name, int control_volume, int* size);
...
```
(my C is rusty and I don't remember if/how aligiment affect "bool" arrays)

prusse-martin (changes requested):

src/alfasim_sdk/alfasim_sdk_api/api.h
prusse-martin: Check existing [error codes](https://github.com/ESSS/alfasim-sdk/blob/dbf8d6513bf544cf4fa683ff01ccd86b15515c9a/src/alfasim_sdk/alfasim_sdk_api/common.h#L9-L20).

Maybe:
- UNDEFINED_DATA for when the user ask for an invalid `prop_name` (btw, are `prop_name` fixed? If so maybe we should have one specific function for each property);
- OUT_OF_BOUNDS for invalid `control_volume` (but arguably UNDEFINED_DATA  also couldbe used I guess);

See BUFFER_SIZE_INSUFFICIENT in the list (with an near 0 number)? This is because it was planned to the buffers to be allocated by the caller.

prusse-martin: Also.
By `wall_names`. Do you mean `material_names`? 

I ask that because a point in a pipe can have just one wall, one wall can have multiple (unnamed layers), and every layer have one material (named).
And a wall could have multiple layers with the same material.

ggrbill: @prusse-martin perhaps we could discuss it in a call. I don't remember the details, but this definition of outputs on API functions was discussed a lot before Tarcisio and William implemented it in this way.

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,51 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
     void* ctx, double* out, const char* var_name, int phase_id
 );
 
+/*!
+    Gets the wall names and wall properties for a given control volume. The wall names will be given
```
ggrbill, line 816: Did you check if this documentation is already being shown in ALFAsim-SDK's docs?
it seems not: https://alfasim-sdk--398.org.readthedocs.build/en/398/plugins/99_1_solver_api_reference.html#alfasim-s-solver-data

You have to go to the docs files and add a line to point to this function. In this way, you will see appearing the documentation.

ggrbill (changes requested):
Nice job!

Please, check the appearance of the new documentation.

src/alfasim_sdk/alfasim_sdk_api/api.h
alexandrebbruno: hmmm, ok

- Regarding the `void* ` for `out` it seems fair to create two separate methods. 

- About the error_code, thanks for the suggestions, I will make those changes (your ideas are much better)

- `prop_name ` can be a set of 7 names, but I think that having one specific function for each property could lead to some code replication. 

> By wall_names. Do you mean material_names?
Actually, it suppose to be `material_names`. I decided to export the names considering that the user could want to know the material's name to take some action (maybe use predefined values for other property). But we can think about create a different method to return that. (`get_wall_material_names`)

prusse-martin: > ... export the names considering that the user could want to know the material's name to take some action...

Those are user inputs not fixed names, they could be in English, Portuguese, Russian, Mandarin...
We should not guess what we are dealing with, forsimulation purposes the name should not be relevant and use just the physical properties of the material.

eniohayashi (approved):

src/alfasim_sdk/alfasim_sdk_api/api.h
alexandrebbruno: I applied all the changes we talked about in the meeting

- [ ] split get_wall_properties into get_wall_properties, get_wall_names and get_wall_check_for_fluid_layer
- [ ] use OUT_OF_BOUNDS instead of UNDEFINED_DATA 

I also checked with alfatwin if get_wall_names is needed. The answer was yes.

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,51 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+    - Dynamic Viscosity
+
+    List of `prop_name`:
+    - `"layer_thickness"`: Wall Layer Thickness [m]
+    - `"rho"`: Wall Density [kg/m3]
+    - `"cp"`: Wall Specific Heat Capacity [J/kg.K]
+    - `"volumetric_thermal_expansion"`: Wall Volumetric Thermal Expansion [1/K]
+    - `"cP"`: Wall dynamic viscosity [cP]
+    - `"k"` : Wall Thermal Conductivity [W/m.degC]
+    - `"layer_is_fluid"`: Informs if the wall is a fluid [1 - is fluid; 0 - is not fluid]
```
alexandrebbruno, line 833: - [ ] remove `layer_is_fluid`

src/alfasim_sdk/alfasim_sdk_api/api.h
alexandrebbruno: done :D

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,104 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+        ctx, &material_names_in_wall, control_volume_id, &size_wall);
+    ~~~~~
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] material_names_in_wall Names of the material presented in the wall.
+    @param[in] control_volume_id Control volume id.
+    @param[out] size Size of the `wall_names` and `prop_values` array of values.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_wall_material_names(void* ctx, char*** wall_names, int control_volume_id, int* size);
```
prusse-martin, line 884:
```suggestion
    @param[in] ctx ALFAsim's plugins context.
    @param[out] material_names_in_wall Names of the material presented in the wall.
    @param[in] control_volume_id Control volume id.
    @param[out] size Size of the `material_names_in_wall` array.
    @return An #error_code value.
*/
DLL_EXPORT int get_wall_material_names(void* ctx, char*** material_names_in_wall, int control_volume_id, int* size);
```

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,104 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+    ~~~~~
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] prop_values Properties in each layer of the wall array.
+    @param[in] prop_name String with the property name. See the list of possible values above.
+    @param[in] control_volume_id Control volume id.
+    @param[out] size Size of the `wall_names` and `prop_values` array of values.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_wall_properties(void* ctx, double** prop_values, const char* prop_name, int control_volume_id, int* size);
```
prusse-martin, line 856:
```suggestion
    @param[in] ctx ALFAsim's plugins context.
    @param[out] prop_values Properties in each layer of the wall array.
    @param[in] prop_name String with the property name. See the list of possible values above.
    @param[in] control_volume_id Control volume id.
    @param[out] size Size of the `prop_values` array.
    @return An #error_code value.
*/
DLL_EXPORT int get_wall_properties(void* ctx, double** prop_values, const char* prop_name, int control_volume_id, int* size);
```

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -812,6 +812,104 @@ DLL_EXPORT int get_liq_liq_flow_pattern_input_variable(
...
+        ctx, &wall_material, control_volume_id, &size_wall);
+    ~~~~~
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[out] wall_material Informs if the wall layer material type [1 - is fluid; 0 - is not fluid]
+    @param[in] control_volume Control volume id.
+    @param[out] size Size of the `wall_names` and `prop_values` array of values.
+    @return An #error_code value.
+*/
+DLL_EXPORT int get_wall_material_type(void* ctx, int** wall_material, int control_volume_id, int* size);
```
prusse-martin, line 911:
```suggestion
    @param[in] ctx ALFAsim's plugins context.
    @param[out] wall_material_type Informs if the wall layer material type [1 - is fluid; 0 - is not fluid]
    @param[in] control_volume Control volume id.
    @param[out] size Size of the `wall_material_type` array.
    @return An #error_code value.
*/
DLL_EXPORT int get_wall_material_type(void* ctx, int** wall_material_type, int control_volume_id, int* size);
```

src/alfasim_sdk/alfasim_sdk_api/detail/api_pointers.h
```diff
@@ -74,6 +74,25 @@ using get_wall_interfaces_temperature_func = int (*)(
...
+    char*** wall_names,
+    int control_volume_id,
+    int* size
+    );
+using get_wall_material_type_func = int (*) (
+    void* ctx,
+    int** wall_material,
+    int control_volume_id,
+    int* size
+    );
```
prusse-martin, line 95:
```suggestion
using get_wall_material_type_func = int (*) (
    void* ctx,
    int** wall_material_type,
    int control_volume_id,
    int* size
    );
```

prusse-martin (approved):
Just some string updates.

ggrbill (approved):

PR #399, opened 2024-08-29
Add logger capability to plugins developers

alexandrebbruno:
PR description:
- It allows the plugin developer to add warning and information messages to the calc solver.

https://esss.atlassian.net/browse/ASIM-5645

### DIFF
--- a/docs/source/plugins/99_1_solver_api_reference.rst
+++ b/docs/source/plugins/99_1_solver_api_reference.rst
@@ -163,6 +163,11 @@ ALFAsim's Solver Data
 
 .. doxygenfunction:: get_flow_pattern
 
+.. doxygenfunction:: set_warning_message
+
+.. doxygenfunction:: set_information_message
+
+
 
 Unit Cell Model (UCM) helpers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/plugins/99_1_solver_api_reference.rst
+++ b/docs/source/plugins/99_1_solver_api_reference.rst
@@ -163,9 +163,9 @@ ALFAsim's Solver Data
 
 .. doxygenfunction:: get_flow_pattern
 
-.. doxygenfunction:: set_warning_message
+.. doxygenfunction:: log_warning_message
 
-.. doxygenfunction:: set_information_message
+.. doxygenfunction:: log_information_message
 
 
 
--- a/docs/source/plugins/99_1_solver_api_reference.rst
+++ b/docs/source/plugins/99_1_solver_api_reference.rst
@@ -163,11 +163,6 @@ ALFAsim's Solver Data
 
 .. doxygenfunction:: get_flow_pattern
 
-.. doxygenfunction:: log_warning_message
-
-.. doxygenfunction:: log_information_message
-
-
 
 Unit Cell Model (UCM) helpers
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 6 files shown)


ThalesCarl (approved):
Nice work :100:

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -620,6 +620,24 @@ DLL_EXPORT int get_deposition_thickness(
...
+DLL_EXPORT int set_warning_message(const char* warning_message);
+
+/*!
+    Sets an information message at the solver logger.
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[in] info_message Information Message.
+    @return An #error_code value.
+*/
+DLL_EXPORT int set_information_message(const char* info_message);
```
prusse-martin, line 639: `set_` seems to imply there is only one of each.
Maybe the prefix should be `add_` or `log_` (with a matching change to the documentation).

Also note that the signature is missing the `ctx` argument.

prusse-martin (changes requested):

docs/source/plugins/99_1_solver_api_reference.rst
```diff
@@ -163,6 +163,11 @@ ALFAsim's Solver Data
 
 .. doxygenfunction:: get_flow_pattern
 
+.. doxygenfunction:: set_warning_message
+
+.. doxygenfunction:: set_information_message
+
```
ggrbill, line 169: I would create a new section in the documentation, something like:: `Solver Logging` or just `Logging` or `Warning and Information Logging`. Also, locate this section right after `Liquid-Liquid Mechanistic Model helpers` Section.

If you think that the section needs a short description, feel free to add one.

src/alfasim_sdk/alfasim_sdk_api/api.h
ggrbill: I agree with both modifications. 
- [ ] change the prefix of the function names to `log_`
- [ ] add `ctx` parameter

ggrbill (changes requested):

caio-ramos (approved):

docs/source/plugins/99_1_solver_api_reference.rst
alexandrebbruno: sure, done

prusse-martin (approved):

src/alfasim_sdk/alfasim_sdk_api/api.h
alexandrebbruno: done :D

src/alfasim_sdk/alfasim_sdk_api/api.h
```diff
@@ -620,6 +620,26 @@ DLL_EXPORT int get_deposition_thickness(
...
+
+    @param[in] ctx ALFAsim's plugins context.
+    @param[in] plugin_id ALFAsim's plugins id
+    @param[in] warning_message Warning Message.
+    @return An #error_code value.
+*/
+DLL_EXPORT int log_warning_message(void* ctx, const char* plugin_id, const char* warning_message);
+
+/*!
+    Logs an information message at the solver logger.
```
ggrbill, line 634: What is the difference between warning and information? When and where is each one shown?
Perhaps, It would be more informative to the plugin dev explaining these differences in the docs.

ggrbill (approved):
Good Job!

src/alfasim_sdk/alfasim_sdk_api/api.h
alexandrebbruno: The Warning logs the message in the following structure: "Warning: " + warning_message while the Information Message logs the information message directly.
I can put that in the doc

eniohayashi (approved):
Nice job!

PR #409, opened 2024-10-01
Improve documentation

eniohayashi:
* Fix documentation for conda env creation
* General documentation review

### DIFF
--- a/docs/source/plugins/01_quick_start.rst
+++ b/docs/source/plugins/01_quick_start.rst
@@ -18,16 +18,15 @@ the official docs <https://www.python.org/downloads/>`_.
 
 .. note::
 
-    It is recommended that you install the |sdk| using the conda package manager. With the commands below, a conda
-    environment would be created with |sdk| installed on it.
+    It is recommended to install the |sdk| using the conda package manager. The following commands will create a conda environment with |sdk| installed.
 
     .. code-block:: bash
 
-        >>> conda env create -n sdk_env
+        >>> conda create -n sdk_env
         >>> conda activate sdk_env
         >>> pip install alfasim-sdk
 
-    Another approach could be adding a ``environment.yml`` to your project and declare the ``alfasim-sdk`` inside the ``pip`` dependencies.
+    Alternatively, an ``environment.yml`` can be added to your project and ``alfasim-sdk`` can be declared at ``pip`` dependencies.
 
     For more details, see `Getting started with conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`_.
 
@@ -100,16 +99,16 @@ After the execution of the :program:`new` command the generated plugin project w
 The highlights here are for:
 
 :plugin.yaml: File with all information about the plugin that will be used by |alfasim|.
-:myplugin.py: Implementation of the hooks for customization of the UI interface, or the solver configuration hooks.
+:myplugin.py: Implementation of the hooks for customization of the user interface or the solver configuration hooks.
 :myplugin.c:  Implementation of the hooks for customization of solver.
 
-Check out the :ref:`Plugin Structure section <plugin_structure-section>` for more details about how the folder and files are structured, and
-also, check the :ref:`plugin-by-example-section` that shows how to create simple plugins that make use of the `User Interface Hooks` and the `Solver Hooks`.
+Refer to the :ref:`Plugin Structure section <plugin_structure-section>` for more details about how the folder and files are structured.
+The section :ref:`plugin-by-example-section` shows how to create simple plugins that make use of the `User Interface Hooks` and the `Solver Hooks`.
 
 Invoke
 ------
 
-From version 0.17.0 onwards, |sdk| started using `pyinvoke`_ tasks in order facilitate and standardize the plugins development process
+Since version 0.17.0, |sdk| started using `pyinvoke`_ tasks in order facilitate and standardize the plugins development process
 using pre-defined tasks. A standard ``tasks.py`` file is generated automatically by the :program:`new` command, which can be customized by the user if needed.
 
 To list the tasks available, use:
@@ -124,8 +123,8 @@ To list the tasks available, use:
 Creating a package
 ------------------
 
-From the root directory of the plugin, execute the command `invoke package`.
-This command will compile your C/C++ implementation and include the shared libraries inside a `artifacts` directory and
+From the plugin root directory, execute the command `invoke package`.
+This command will compile your C/C++ implementation and include the shared libraries inside an `artifacts` directory and
 the generated plugin on the root directory with the extension `hmplugin`.
 
 .. code-block:: bash
@@ -171,18 +170,18 @@ Installing the plugin
 ---------------------
 
 With :file:`myplugin.hmplugin` file, it is now possible to install it on |alfasim| application.
-From the menu bar select the `Plugin Manager` option, as illustrated in the figure below:
+At the menu bar select the `Plugin Manager` option, as illustrated in the figure below:
 
 .. image:: /_static/images/quick_start/menu_bar.png
 
-In the `Plugin Manager` windows, install :file:`myplugin.hmplugin` file plugin.
+In the `Plugin Manager` window, install :file:`myplugin.hmplugin` file plugin.
 
 .. image:: /_static/images/quick_start/plugin_manager_empty.png
 
-By clicking on the plugin name, it is possible to visualize the content from the :file:`README.md` file.
+By clicking on the plugin name, it is possible to visualize the content of the :file:`README.md` file.
 
 .. image:: /_static/images/quick_start/plugin_manager_with_plugin.png
 
-Restart the application and check the recently installed plugin available over the :guilabel:`Tree`.
+Restart the application and check the recently installed plugin available at the :guilabel:`Tree`.
 
 .. image:: /_static/images/quick_start/tree_with_plugin.png
--- a/docs/source/plugins/03_plugin_structure.rst
+++ b/docs/source/plugins/03_plugin_structure.rst
@@ -3,7 +3,7 @@
 Plugins Structure
 =================
 
-As mentioned in :ref:`quick-start-section`, the |sdk| package has some utilities to help in the process to create
+As mentioned in :ref:`quick-start-section`, the |sdk| package has some utilities to help in the process of creating
 a plugin file.
 
 First, to create the plugin use :program:`new`:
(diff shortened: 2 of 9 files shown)


docs/source/plugins/06_solver_hooks.rst
```diff
@@ -219,7 +219,7 @@ three-phase flow it is required a plugin that handles the oil-water system.
 
 .. note ::
```
eniohayashi, line 220: @ggrbill  I didn't get this note. Could you help me, pls?

docs/source/plugins/01_quick_start.rst
```diff
@@ -18,16 +18,15 @@ the official docs <https://www.python.org/downloads/>`_.
...
 
     .. code-block:: bash
 
-        >>> conda env create -n sdk_env
+        >>> conda create -n sdk_env
         >>> conda activate sdk_env
         >>> pip install alfasim-sdk
 
-    Another approach could be adding a ``environment.yml`` to your project and declare the ``alfasim-sdk`` inside the ``pip`` dependencies.
+    Alternatively, ``environment.yml`` can be added to your project and ``alfasim-sdk`` can be declared at ``pip`` dependencies.
```
ThalesCarl, line 29:
```suggestion
    Alternatively, an ``environment.yml`` can be added to your project and ``alfasim-sdk`` can be declared at ``pip`` dependencies.
```

docs/source/plugins/01_quick_start.rst
```diff
@@ -100,16 +99,16 @@ After the execution of the :program:`new` command the generated plugin project w
 The highlights here are for:
 
 :plugin.yaml: File with all information about the plugin that will be used by |alfasim|.
-:myplugin.py: Implementation of the hooks for customization of the UI interface, or the solver configuration hooks.
+:myplugin.py: Implementation of the hooks for customization of the user interface, or the solver configuration hooks.
```
ThalesCarl, line 102:
```suggestion
:myplugin.py: Implementation of the hooks for customization of the user interface or the solver configuration hooks.
```

docs/source/plugins/03_plugin_structure.rst
```diff
@@ -21,9 +21,8 @@ First, to create the plugin use :program:`new`:
 Pyinvoke tasks
 --------------
 
-Except for creating the plugin structure, all other activities that you might want to perform, you will use ``pyinvoke`` tasks.
+Except for creating the plugin structure, all other activities that you might want to perform uses the following ``pyinvoke`` tasks:
 
-Below there is detailed list on what they are meant to do. However, in general, the tasks that you will use the most are ``compile``,
 ``package`` and ``update``.
```
ThalesCarl, line 26: I don't get why we need to remove the last line. You can check that this line fits when completed with the tasks description [in the generated documentation](https://alfasim-sdk.readthedocs.io/en/latest/plugins/03_plugin_structure.html).

However, if you still want to remove it, you need to remove the following line as well

```rst
``package`` and ``update``
```

because it seems to be a leftover

docs/source/plugins/04_application_hooks.rst
```diff
@@ -7,8 +7,7 @@ The application hooks allow plugins to add custom models or custom checks in the
 To add an application hook it is necessary to implement the given hook in a python file that is
 already available on your plugin project folder.
 
-As an example, if the ``alfasim-sdk new`` command was created with the name ``myplugin``, the necessary file to
-be customized will be located on: :menuselection:`myplugin --> src --> python --> myplugin.py`
+For example, executing the command ``alfasim-sdk new myplugin``, the file `myplugin.py`, located at :menuselection:`myplugin --> src --> python should be customized.
```
ThalesCarl, line 10:
```suggestion
For example, after executing the command ``alfasim-sdk new myplugin``, the file `myplugin.py`, located at :menuselection:`myplugin --> src --> python, should be customized.
```

docs/source/plugins/05_solver_config_hooks.rst
```diff
@@ -33,7 +33,7 @@ the hook listed below, the plugin can:
  - Add/update layers
 
 .. note::
-    For each new added ``field`` is considered a mass conservation equation and for each new added ``layer`` is considered
+    For each new added ``field`` is considered a mass conservation equation and for each new added ``layer`` are considered
     a momentum conservation and an energy conservation equations, depending on the energy model used at |alfasim|.
```
ThalesCarl, line 37: This phrase is still ambiguous. A suggestion could be:

```suggestion
    For each new added ``field``, an extra mass conservation equation is considered. For each new added ``layer``, an extra momentum and energy conservation equations are considered, depending on the energy model used at |alfasim|.
    
```

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -3,10 +3,10 @@
...
 ============
 
-The present section describes all solver `Hooks` available on |alfasim| plugin infrastructure.
-Solver `Hooks` are |alfasim|'s pre-defined functions that allows the plugin developer to modify/extend |alfasim|'s Solver.
-As already informed in :ref:`Quick Start <quick-start-section>` section once created a plugin using ``new`` option
-on |sdk|'s CLI a new file named ``plugin.c`` will be available to implement those `hooks`.
+The present section describes all solver `Hooks` available in |alfasim| plugin infrastructure.
+Solver `Hooks` are |alfasim|'s pre-defined functions that allow plugin developer to modify or extend |alfasim|'s Solver.
+As already informed in :ref:`Quick Start <quick-start-section>` section, once created a plugin using ``new`` option
+in |sdk|'s CLI a new file named ``plugin.c`` will be available to implement those `hooks`.
```
ThalesCarl, line 9: This seems to be not up to date. Currently, the file created is `plugin.cpp` and not `plugin.c`. Or am I missing something?

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -72,22 +71,23 @@ Update plugin variables
 
 The `hooks` described in this section are related to plugin secondary variables that were registered in the python config
 file, as already explained in :ref:`solver_customization` section. They are called `secondary variables` because they
-are not obtained from |alfasim|'s Solver, these ones are called primary variables and examples of those variables are `pressure`,
-`temperature`, `volume fractions` and `velocities`.
+are not obtained from |alfasim|'s Solver like the `pressure`, `temperature`, `volume fractions` and `velocities`, called primary variables istead.
```
ThalesCarl, line 74:
```suggestion
are not obtained from |alfasim|'s Solver like the `pressure`, `temperature`, `volume fractions` and `velocities`, called primary variables instead.
```

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -117,7 +117,7 @@ This `hook` is used to update plugin's variables that depends on Tracer's mass f
 Source Terms
 ------------
 
-The `hooks` showed in this section can be considered as the most important. Since they allow the plugin to change the
+The `hooks` showed in this section can be considered as the most important since they allow the plugin to change the
```
ThalesCarl, line 120:
```suggestion
The `hooks` showed in this section can be considered as the most important hooks since they allow the plugin to change the
```

docs/source/plugins/06_solver_hooks.rst
```diff
@@ -174,8 +174,8 @@ Internal deposit layer
 ----------------------
 
 When a new phase is added to the hydrodynamic model using the :py:class:`~alfasim_sdk.AddPhase` type, it is possible
-to consider the deposition inside the pipeline walls. If so, calculate the thickness of the deposited layer on a given
-phase through the usage of :py:func:`~alfasim_sdk._internal.hook_specs.update_internal_deposition_layer`. By default, the layer
+to consider the deposition inside the pipeline walls. In this case, the thickness of the deposited layer on a given
+phase is calculated through the usage of :py:func:`~alfasim_sdk._internal.hook_specs.update_internal_deposition_layer`. By default, the layer
```
ThalesCarl, line 178:
```suggestion
to consider the deposition inside the pipeline walls. In this case, the thickness of the deposited layer on a given
phase should be calculated through the usage of :py:func:`~alfasim_sdk._internal.hook_specs.update_internal_deposition_layer`. By default, the layer
```

Proposing this change, because `is calculated` gives the idea that the deposition layer is calculated by alfasim instead of the user calculating it and imposing the value in the hook

ThalesCarl (approved):
Nice refactoring of the documentation.

docs/source/plugins/06_solver_hooks.rst
ggrbill: This is tricky to explain.

I think we could change the name of things. Like the `Two-phase (Gas-Liquid) System of Mechanistic Model` Could be just `Two-Phase Mechanistic Model. 

Then, this note could be:
"
Note that the two-phase (Gas-Liquid) Mechanistic Model can be used for both two and three-phase flows (Fluid FLow Type selected by the user in ALFAsim). When no Liquid-Liquid System is implemented by the plugin, the three-phase flow is adapted to work with the traditional two-phase Mechanistic Model, in which all liquid phases are considered as only one.
"

See if you understand it. Perhaps it needs a little bit more improvement.

ggrbill (approved):
Very nice review!

docs/source/plugins/06_solver_hooks.rst
eniohayashi: I see that this is hard to explain in this context since it requires a deeper knowledge of the hydrodynamic modeling. But I think that your suggestion is much better than the original text. Thanks!

docs/source/plugins/01_quick_start.rst
eniohayashi: Done.

docs/source/plugins/01_quick_start.rst
eniohayashi: Done.

docs/source/plugins/03_plugin_structure.rst
eniohayashi: You're right. It was a leftover, but I see that it could be kept as well.  I rewrote both paragraphs as 
```
Once the plugin structure is created, you can perform all necessary activities using ``pyinvoke`` tasks, with ``compile``, ``package``  and ``update`` beeing the 
most commonly used during plugin development. Below you can find a list of all available along with their descriptions.
```

docs/source/plugins/04_application_hooks.rst
eniohayashi: Done.

docs/source/plugins/05_solver_config_hooks.rst
eniohayashi: Done.

docs/source/plugins/06_solver_hooks.rst
eniohayashi: You're right. Fixed.

docs/source/plugins/06_solver_hooks.rst
eniohayashi: Done.

docs/source/plugins/06_solver_hooks.rst
eniohayashi: I think I'll keep it as it is because using `hooks` again sounds redundant.

docs/source/plugins/06_solver_hooks.rst
eniohayashi: I agree! Done.

PR #411, opened 2024-11-22
Fix HM results reader

ggrbill:
### ASIM-5935: Fix History Matching results reader
- Since the shape of data is different than expected, the reader had to be updated to consider the correct shape of data.

### DIFF
--- a/tests/results/test_aggregator.py
+++ b/tests/results/test_aggregator.py
@@ -511,13 +511,18 @@ def test_read_history_matching_result_data(
     Check reading the result of both HM type analysis. Both results are available simultaneously by
     the means of the fixtures, but only one is used at a time.
     """
+    import numpy as np
+
     # Setup.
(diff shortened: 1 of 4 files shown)


BeneBr (approved):

tests/results/test_aggregator.py
```diff
@@ -511,13 +513,19 @@ def test_read_history_matching_result_data(
     Check reading the result of both HM type analysis. Both results are available simultaneously by
     the means of the fixtures, but only one is used at a time.
     """
+    import numpy as np
+
     # Setup.
+    slicer: Callable[[int], Any]
```
ro-oliveira95, line 519: This var can be removed now.

ro-oliveira95 (approved):

prusse-martin:
Don't forget to update the change log.

prusse-martin (approved):

PR #412, opened 2024-12-03
[ASIM-5943] Add FluidMaterialConvectionCorrelation

BeneBr:
Add a new transfer mechanism for fluid materials composed by a parcel of radiation and convection.

ASIM-5943

### DIFF
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -539,3 +539,14 @@ class ForchheimerCoefficientsOption(Enum):
 
     ReservoirParameters = "reservoir_parameters"
     FlowCoefficients = "flow_coefficients"
+
+class FluidMaterialConvectionCorrelation(Enum):
+    """
+    The heat transfer mechanism for fluid materials composed by a parcel of radiation and
+    convection.
+    """
+
+    HasanKabir1994 = "Hasan and Kabir 1994"
+    HasanKabir2002 = "Hasan and Kabir 2002"
+    FuzongZhou2013 = "Fuzong Zhou 2013"
+    Holman1981 = "Holman 1981"
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -540,6 +540,7 @@ class ForchheimerCoefficientsOption(Enum):
     ReservoirParameters = "reservoir_parameters"
     FlowCoefficients = "flow_coefficients"
 
+
 class FluidMaterialConvectionCorrelation(Enum):
     """
     The heat transfer mechanism for fluid materials composed by a parcel of radiation and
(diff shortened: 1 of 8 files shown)


nicoddemus (changes requested):
LGTM, but is missing a CHANGELOG entry. :+1:

caio-ramos (approved):

src/alfasim_sdk/_internal/constants.py
```diff
@@ -539,3 +539,15 @@ class ForchheimerCoefficientsOption(Enum):
...
+class FluidMaterialConvectionCorrelation(Enum):
+    """
+    The heat transfer mechanism for fluid materials composed by a parcel of radiation and
+    convection.
+    """
+
+    HasanKabir1994 = "Hasan and Kabir 1994"
+    HasanKabir2002 = "Hasan and Kabir 2002"
+    FuzongZhou2013 = "Fuzong Zhou 2013"
+    Holman1981 = "Holman 1981"
```
prusse-martin, line 553: Not a change request.

I just noted we have some enums that appear to make reference to some papers.
What if some time we add a more comprehensive documentation with better references to the respective papers so it ends up in the online documentation?
Or is that just overkill?

@nicoddemus @eniohayashi 
What are your thoughts about this?

prusse-martin (approved):

src/alfasim_sdk/_internal/constants.py
caio-ramos: In general, the complete reference is added to the calc function.

nicoddemus: > In general, the complete reference is added to the calc function.

Probably a good idea to add here as well, because IIUC the calc functions are not visible to the users.

nicoddemus (approved):

src/alfasim_sdk/_internal/constants.py
eniohayashi: Right, calc documentation is only for dev purpose. Maybe a tooltip with a brief description would be good. I think @caio-ramos could add it here later.

eniohayashi (approved):

src/alfasim_sdk/_internal/constants.py
caio-ramos: Actually, I already suggested to @BeneBr to add a tooltip with the following information: 
```
The heat transfer mechanism for fluid materials is composed of a parcel of radiation and convection. The following correlations are available to model de convection term:
- Hasan, A. R., and C. S. Kabir. "Aspects of Wellbore Heat Transfer During Two-Phase Flow." SPE Prod & Fac 9 (1994): 211–216.
- Hasan, A. R., and C. S. Kabir. "Fluid Flow and Heat Transfers in Wellbores." Society of Petroleum Engineers (2002).
- Zhou, Fuzong. "Research on heat transfer in geothermal wellbore and surroundings." PhD Thesis. Institut für Angewandte Geowissenschaften (2013).
- Holman, J.P. "Heat transfer". McGraw-Hill (1981), New York. 
```

nicoddemus: Folks let's not mix things, one thing is the GUI and the other is the SDK.

I'm talking about adding a link here to the documentation for the SDK user, which is independent from adding the same information to the GUI. I think we should have both.

eniohayashi: Oh, ok. My bad here. I agree, we should have both.

PR #414, opened 2024-12-18
Add support for Results API

ro-oliveira95:
Add support for the ALFAsim's server Results API while also improving the way users can interact with the UQ results.

ASIM-5920

### DIFF
--- a/src/alfasim_sdk/result_reader/aggregator.py
+++ b/src/alfasim_sdk/result_reader/aggregator.py
@@ -103,12 +103,13 @@ A map of base time steps to the metadata contained in the respective result file
 All the metadata will represents the same kind of output (profiles/trends).
 """
 
-HistoryMatchingResultKeyType = str
-"""\
-A HM result key is simply the id of the parametric var associated with that particular result.
-"""
 
-TimeSetInfoItem = namedtuple("TimeSetInfoItem", "global_start size uuid")
+class TimeSetInfoItem(NamedTuple):
+    global_start: TimeStepIndex
+    size: int
+    uuid: str
+
+
 TimeSetInfo = Dict[int, TimeSetInfoItem]
 
 
@@ -383,13 +384,25 @@ class HistoricDataCurveMetadata:
 
 
 @dataclasses.dataclass(frozen=True)
+class HMOutputKey:
+    parametric_var_id: str
+
+    def __str__(self):
+        return self.parametric_var_id
+
+    @classmethod
+    def from_string(cls, raw_output_key: str) -> Self:
+        return cls(raw_output_key)
+
+
+@dataclasses.dataclass(frozen=True)
 class HistoryMatchingMetadata:
     """
     Holder for the History Matching results metadata.
     """
 
     #: Map of the data id and its associated metadata.
-    hm_items: Dict[str, HMItem]
+    hm_items: Dict[HMOutputKey, HMItem]
     #: Map of observed curve id to a dict of Quantity of Interest data, populated with keys
     #: 'trend_id' and 'property_id'. This represents the setup for this HM analysis.
     objective_functions: Dict[str, Dict[str, str]]
@@ -435,16 +448,7 @@ class HistoryMatchingMetadata:
             )
 
     @classmethod
-    def empty(cls, result_directory: Path) -> Self:
-        return cls(
-            hm_items={},
-            objective_functions={},
-            parametric_vars={},
-            result_directory=result_directory,
-        )
-
-    @classmethod
-    def from_result_directory(cls, result_directory: Path) -> Self:
+    def from_result_directory(cls, result_directory: Path) -> Self | None:
         """
         Read History Matching results metadata from result directory/file.
 
@@ -453,9 +457,11 @@ class HistoryMatchingMetadata:
 
         def map_meta_items(
             hm_metadata: Dict,
-        ) -> Dict[str, HistoryMatchingMetadata.HMItem]:
+        ) -> Dict[HMOutputKey, HistoryMatchingMetadata.HMItem]:
             return {
-                key: HistoryMatchingMetadata.HMItem.from_dict(data)
+                HMOutputKey.from_string(key): HistoryMatchingMetadata.HMItem.from_dict(
+                    data
+                )
                 for key, data in hm_metadata.items()
             }
 
@@ -466,15 +472,12 @@ class HistoryMatchingMetadata:
 
         with open_result_file(result_directory) as result_file:
             if not result_file:
-                return cls.empty(result_directory=result_directory)
+                return None
 
             loaded_metadata = json.loads(
                 result_file[META_GROUP_NAME].attrs[HISTORY_MATCHING_GROUP_NAME]
             )
-
-            if len(loaded_metadata) == 0:
-                return cls.empty(result_directory=result_directory)
-
+            assert len(loaded_metadata) > 0
             some_item_metadata = list(loaded_metadata.values())[0]
 
             objective_functions = some_item_metadata["objective_functions"]
@@ -1947,7 +1950,9 @@ def read_global_sensitivity_coefficients(
     return result
 
 
-def read_history_matching_metadata(result_directory: Path) -> HistoryMatchingMetadata:
+def read_history_matching_metadata(
+    result_directory: Path,
+) -> HistoryMatchingMetadata | None:
     """
     :param result_directory:
         The directory to lookup for the History Matching result file.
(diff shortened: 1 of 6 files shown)


src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -149,28 +157,38 @@ class BaseUQMetaData:
...
+    property_name: str
+    element_name: str
+
+    def __str__(self):
+        return f"{self.property_name}@{self.element_name}"
+
+    @classmethod
+    def from_string(cls, raw_output_key: str) -> Self:
+        r = re.match(r"^(.*)@(.*)", raw_output_key)
+        assert r is not None, f"Invalid output key: {raw_output_key}"
```
nicoddemus, line 180: This should be a normal exception (ValueError maybe?), not an assert.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -149,28 +157,38 @@ class BaseUQMetaData:
...
+    category: str
+    network_element_name: Optional[str]
+    position: Optional[float]
+    position_unit: Optional[str]
+    unit: str
 
 
-@attr.s(slots=True, hash=False)
+@dataclasses.dataclass(slots=True, frozen=True)
+class UPOutputKey:
```
nicoddemus, line 170: Great abstraction!

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -149,28 +157,38 @@ class BaseUQMetaData:
...
+
+    @classmethod
+    def from_string(cls, raw_output_key: str) -> Self:
+        r = re.match(r"^(.*)@(.*)", raw_output_key)
+        assert r is not None, f"Invalid output key: {raw_output_key}"
+        property_name, element_name = r.groups()
+        return cls(property_name, element_name)
+
+
+@dataclasses.dataclass(slots=True, frozen=True)
```
nicoddemus, line 185: This used `hash=False` previously... not sure why though, making it frozen seems reasonable.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -231,13 +243,30 @@ def map_data(
...
+    parametric_var_id: str
+    element_name: str
+
+    def __str__(self):
+        return f"{self.property_name}::{self.parametric_var_id}@{self.element_name}"
+
+    @classmethod
+    def from_string(cls, raw_output_key: str) -> Self:
+        r = re.match(r"^(.*)::(.*)@(.*)", raw_output_key)
+        assert r is not None, f"Invalid output key: {raw_output_key}"
```
nicoddemus, line 258: Same about this being a proper exception.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -525,14 +616,18 @@ class ALFASimResultMetadata:
...
+    profiles: dict[OutputKeyType, ProfileMetaItem]
+    trends: dict[OutputKeyType, TrendMetaItem]
+    time_sets: list[SourceTimeSetKeyType]
+    time_sets_unit: str
+    time_steps_boundaries: tuple[
+        tuple[TimeStepIndex, TimeStepIndex], tuple[TimeStepIndex, TimeStepIndex]
+    ]
+    time_set_info: dict[
+        Literal["profiles", "trends"], dict[TimeStepIndex, TimeSetInfoItem]
+    ]  # TODO: verify if keys are TimeStep.
```
nicoddemus, line 628: TODO

nicoddemus (commented):
Great work!

tests/results/test_aggregator.py
```diff
@@ -210,7 +209,9 @@ def test_concatenate_metadata_handle_time_sets(results: Results) -> None:
 
 def test_concatenate_metadata_error_conditions_directory(results: Results) -> None:
     md = results.metadata
-    other_md = attr.evolve(md, result_directory=md.result_directory / "other")
+    # other_md = attr.evolve(md, result_directory=md.result_directory / "other")
```
ro-oliveira95, line 212: - [ ] Remove left over

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -204,22 +220,18 @@ def from_dict(cls, data: Dict[str, Any]) -> Self:
...
             )
 
-    items: Dict[str, UPItem] = attr.ib(validator=attr.validators.instance_of(Dict))
-    result_directory: Path = attr.ib(validator=attr.validators.instance_of(Path))
-
-    @classmethod
-    def empty(cls, result_directory: Path) -> Self:
-        return UncertaintyPropagationAnalysesMetaData(
-            items={}, result_directory=result_directory
-        )
```
ro-oliveira95, line 214: A note for reviewers: this was removed there should be no case in which we have empty metadata for this type of results, once they don't have the concept of partial completion. The presence of these empty objects could lead to confusion and ultimately bugs, so they were ripped of.

src/alfasim_sdk/result_reader/aggregator.py
ro-oliveira95: Yeah I'm also not sure why, but at least for our internal use cases frozen works fine so I guess we're good.

src/alfasim_sdk/result_reader/aggregator.py
ro-oliveira95: - [ ] Remove TODO

I would ask @prusse-martin on this one but he is on vacation right now. I will leave this as `TimeStepIndex` for now but will ask him to do a post-mortem review on this and in the `Trend/ProfileMetaItem` definitions, then any changes can be addressed in a follow-up PR.

src/alfasim_sdk/result_reader/reader.py
```diff
@@ -295,3 +330,154 @@ def get_profile_metadata(meta: Dict) -> ProfileMetadata:
             get_profile_metadata(profile_metadata)
             for profile_metadata in metadata.profiles.values()
         ]
+
+
+def _non_empty_dict_validator(values_type: type) -> Callable:
+    def validator(inst: Any, attribute: attr.Attribute, value: Any) -> None:
```
cauebs, line 336: You've probably thought about this already, but it would be good if these had more specific types. Not sure if `object` instead of `Any` would work, but at least `Callable[[Any, attr.Attribute, Any] None]`?

cauebs: If this is actionable, it applies to the other validators, of course.

cauebs (approved):

src/alfasim_sdk/result_reader/reader.py
```diff
@@ -295,3 +330,154 @@ def get_profile_metadata(meta: Dict) -> ProfileMetadata:
...
+
+        return cls(
+            timeset=read_uq_time_set(
+                result_dir, GLOBAL_SENSITIVITY_ANALYSIS_GROUP_NAME
+            ),
+            coefficients=read_global_sensitivity_coefficients(metadata),
+            metadata=metadata,
+        )
+
+    def get_sensibility_curve(
```
BeneBr, line 382: All this feature is called ```Global Sensitivity Analysis```, i believe that the correct name for this functions is:
```suggestion
    def get_sensitivity_curve(
```

tests/results/test_result_reader.py
```diff
@@ -134,3 +152,185 @@ def test_metadata_string():
 
     global_trend = GlobalTrendMetadata("property")
     assert str(global_trend) == "property"
+
+
+def test_global_sensibility_analysis_results_reader(
```
BeneBr, line 157:
```suggestion
def test_global_sensitivity_analysis_results_reader(
```

BeneBr (approved):

src/alfasim_sdk/result_reader/reader.py
ro-oliveira95: Although "sensitivity" and "sensibility" convey the same meaning, I agree that it's better to be consistent. Will change that, good catch.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -93,19 +100,20 @@
...
-"""
 
-TimeSetInfoItem = namedtuple("TimeSetInfoItem", "global_start size uuid")
+class TimeSetInfoItem(NamedTuple):
+    global_start: TimeStepIndex
+    size: int
+    uuid: str
+
+
 TimeSetInfo = Dict[int, TimeSetInfoItem]
```
prusse-martin, line 117: The key here could be `TimeStepIndex`.

src/alfasim_sdk/result_reader/aggregator.py
```diff
@@ -923,7 +1020,7 @@ def read_time_set_info(
         if size > 0:
             limits[key] = TimeSetLimits(dataset[0], dataset[-1], dataset)
         time_set_info[key] = TimeSetInfoItem(
-            global_start=start, size=size, uuid=time_set_uuid
+            global_start=int(start), size=int(size), uuid=time_set_uuid
```
prusse-martin, line 1023: Do you remember why did you need to type cast those arguments (start and size)?

ro-oliveira95: IIRC they were being assigned to numpy's type `np.uint8` and the (de)serialization engine (`pydantic` or `pyserde`, not sure which one) didn't considered it as a valid type for that particular field.

PR #417, opened 2025-01-13
Add equality checks to the UQ result readers.

ro-oliveira95:
ASIM-5939

### DIFF
--- a/src/alfasim_sdk/result_reader/reader.py
+++ b/src/alfasim_sdk/result_reader/reader.py
@@ -355,7 +355,7 @@ def _non_empty_attr_validator(attr_name: str) -> Callable:
     return validator
 
 
-@define(frozen=True)
+@define(frozen=True, eq=False)
 class GlobalSensitivityAnalysisResults:
     timeset: np.ndarray = attr.field(validator=attr.validators.min_len(1))
     coefficients: dict[GSAOutputKey, np.ndarray] = attr.field(
@@ -393,6 +393,22 @@ class GlobalSensitivityAnalysisResults:
         domain = Array(self.timeset, "s")
         return Curve(image=image, domain=domain)
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, GlobalSensitivityAnalysisResults):
+            return False
+
+        return (
+            np.array_equal(self.timeset, other.timeset)
+            and _all_dict_close(self.coefficients, other.coefficients)
+            and self.metadata == other.metadata
+        )
+
+
+def _all_dict_close(a: dict[Any, np.ndarray], b: dict[Any, np.ndarray]) -> bool:
+    if a.keys() != b.keys():
+        return False
+    return all(np.array_equal(a[key], b[key]) for key in a)
+
 
 @define(frozen=True)
 class _BaseHistoryMatchingResults:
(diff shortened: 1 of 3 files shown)


src/alfasim_sdk/result_reader/reader.py
```diff
@@ -443,6 +459,18 @@ def from_directory(cls, result_dir: Path) -> Self | None:
             metadata=metadata,
         )
 
+    def __eq__(self, other: Any) -> bool:
+        if not isinstance(other, HistoryMatchingProbabilisticResults):
+            return False
+
```
nicoddemus, line 465: Can you use `reveal_type(other)` here to ensure it is `HistoryMatchingProbabilisticResults`?

If it is not, try using `other: object` in the signature.

tests/results/test_result_reader.py
```diff
@@ -188,6 +189,11 @@ def test_global_sensitivity_analysis_results_reader(
         qoi_data_index=0,
     )
 
+    # Test equality check.
+    assert results == results
```
nicoddemus, line 193: Just to make sure they are different instances:

```suggestion
    assert results == attr.evolve(results)
```

tests/results/test_result_reader.py
```diff
@@ -204,6 +210,13 @@ def test_deterministic_reader(self, hm_deterministic_results_dir: Path) -> None:
         }
         self._validate_meta_and_historic_curves(results)
 
+        # Test equality check.
+        assert results == results
```
nicoddemus, line 214:
```suggestion
        assert results == attr.evolve(results)
```

tests/results/test_result_reader.py
```diff
@@ -217,6 +230,16 @@ def test_probabilistic_reader(self, hm_probabilistic_results_dir: Path) -> None:
         )
         self._validate_meta_and_historic_curves(results)
 
+        # Test equality check.
+        assert results == results
```
nicoddemus, line 234:
```suggestion
        assert results == attr.evolve(results)
```

tests/results/test_result_reader.py
```diff
@@ -331,6 +354,11 @@ def test_uncertainty_propagation_results_reader(up_results_dir: Path) -> None:
         sample_indexes=[[0, 0], [0, 1], [0, 2], [0, 3], [0, 4]],
     )
 
+    # Test equality check.
+    assert reader == reader
```
nicoddemus, line 358:
```suggestion
    assert reader == attr.evolve(reader)
```

nicoddemus (approved):

src/alfasim_sdk/result_reader/reader.py
ro-oliveira95: It is indeed `HistoryMatchingProbabilisticResults`, so I think we're good.

tests/results/test_result_reader.py
nicoddemus: To elaborate, if you don't implement `__eq__` at all it will compare by identity, and the tests as written previously would pass:

```python
>>> class C:
...     def __init__(self, x): self.x=x
... 
>>> c = C(10)
>>> c == c
True
>>> c != object()
True
>>> c != C(20)
True
>>> c == C(10)
False
```

Note that I did not implement `__eq__` at all -- comparing with a copy ensures that `==` is actually correct (the last check is against a copy which is equal but does not compare equal because `__eq__` was not implemented).

TLDR always test `__eq__` against copies.

prusse-martin: > ...  if you don't implement __eq__ at all it will compare by identity...

@nicoddemus forgot to mention that it is this way be cause `object` implement `__eq__` by identity.
`__eq__` can be overwritten by any class an will be inherited.

BeneBr (approved):

PR #422, opened 2025-01-31
[ASIM-5869] - Add water specific enthalpy for constant PVT

BeneBr:
# Description
Add the water specific enthalpy property for constant PVT description.

### DIFF
--- a/src/alfasim_sdk/alfasim_sdk_api/api.h
+++ b/src/alfasim_sdk/alfasim_sdk_api/api.h
@@ -693,7 +693,13 @@ DLL_EXPORT int get_tracer_name(void* ctx, char* out, void* reference, int size);
 
 /*!
     Gets the tracer reference for a given tracer name. This function is important to obtain the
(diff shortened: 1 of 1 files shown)


nicoddemus (approved):
Changelog?

eniohayashi (approved):

caio-ramos:
Actually, following the procedure defined in [`create_constant`](https://github.com/ESSS/alfasim-sdk/blob/master/src/alfasim_sdk/_internal/alfacase/case_description.py#L2745), the water enthalpy is a function of the water density and water thermal capacity, which are parameters already given by the user. Thus, if we want to keep the previous behavior of `create_constant`, the user should not give such a property.

BeneBr:
Base in the Caio's comment: https://github.com/ESSS/alfasim-sdk/pull/422#issuecomment-2630864035. This property is no longer needed to be added. This will be calculated internally by calc. So i will close this PR.

PR #423, opened 2025-02-04
Migrate to ruff

nicoddemus:
As https://github.com/ESSS/alfasim-sdk/pull/360 shows, black and reorder-python-imports conflict on how they format imports.

Switch over to `ruff` + its builtin import sorter (isort).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,3 +4,30 @@ requires = [
   "wheel",
 ]
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+target-version = "py310"
+src = [
+    "src",
+    "tests",
+]
+format.docstring-code-format = true
+lint.select = ['E4', 'E7', 'E9', 'F', 'I']
+lint.ignore = [
+    "E402",
+    "E711",
+    "E712",
+    "E721",
+    "E722",
+    "E731",
+    "E741",
+    "F401",
+    "F403",
+    "F405",
+    "F507",
+    "F601",
+    "F722",
+    "F811",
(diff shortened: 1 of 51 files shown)


pyproject.toml
```diff
@@ -4,3 +4,30 @@ requires = [
...
 build-backend = "setuptools.build_meta"
+
+[tool.ruff]
+target-version = "py310"
+src = [
+    "src",
+    "tests",
+]
+format.docstring-code-format = true
+lint.select = ['E4', 'E7', 'E9', 'F', 'I']
```
nicoddemus, line 15: Those linter rules were copied from `etk`, where we have already done a good amount of work of selecting good rules that work for us.

ro-oliveira95 (approved):
Nice!

src/alfasim_sdk/_internal/hook_specs_gui.py
```diff
@@ -194,12 +202,11 @@ def alfasim_get_status(
...
         from alfasim_sdk import ErrorMessage
 
+
         # Define MyModel used in this plugin
         @data_model(icon="", caption="My Plugin Model")
         class MyModel:
-            distance = Quantity(
-                value=1, unit="m", caption="Distance"
-            )
+            distance = Quantity(value=1, unit="m", caption="Distance")
```
prusse-martin, line 209: I just ask myself why did it reformat the string.

nicoddemus: It also formats code in docstrings IIRC.

PR #429, opened 2025-02-10
Bring the the release 2024.3.1 changes to master

BeneBr:


src/alfasim_sdk/_internal/version.py
```diff
@@ -1 +1 @@
-__version__ = "1.0.0.dev"
+__version__ = "1.0.0"
```
nicoddemus, line 1: We will need to bump the SDK to `1.1.0.dev`, you will do this after this merge?

nicoddemus (approved):

ro-oliveira95 (approved):

src/alfasim_sdk/_internal/version.py
BeneBr: Yes i will do this after this merge.

PR #443, opened 2025-04-16
Remove results dir from metadata objects

ro-oliveira95:
With the introduction of the ALFAsim client-server a result can be stored in multiples places (and eventually in multiple data stores), so the presence of this info along with the real result metadata doesn't make sense anymore.

ASIM-6121


nicoddemus (approved):
Please update the CHANGELOG, this is a breaking change.

PR #456, opened 2025-06-09
[ASIM] - Add requirements for new plugins

BeneBr:


src/alfasim_sdk/_internal/alfasim_sdk_utils.py
```diff
@@ -92,3 +93,15 @@ def get_extras_default_required_version() -> str:
...
+def get_required_python_version() -> str:
+    """
+    Return the required python version.
+    """
+    major, minor, micro = (
+        sys.version_info.major,
+        sys.version_info.minor,
+        sys.version_info.micro,
+    )
+    return f">={major}.{minor}.{micro}"
```
nicoddemus, line 107: I don't think we need to include the micro version in the required version. Also, to be safe in case the plugin contains compiled code, better include an upper bound to the Python version:


```suggestion
def get_required_python_version() -> str:
    """
    Return the required python version to be included in plugins metadata when using
    the CLI to generate a template.
    """
    # Assume most plugins will use compiled code and include an upper bound.
    major, minor, *_ = sys.version_info
    return f">={major}.{minor},<{major}.{minor + 1}"
```

src/alfasim_sdk/_internal/cli.py
```diff
@@ -104,6 +109,11 @@ def new(dst, caption, plugin_id, author_name, author_email):
         "}",
     ]
 
+    requirements = {
+        "sdk_version": get_extras_default_required_version(),
```
nicoddemus, line 113: Let's rename `get_extras_default_required_version` to better reflect what it does, and for consistency with `get_required_python_version`:

```suggestion
        "sdk_version": get_required_sdk_version(),
```

nicoddemus (approved):

cauebs (approved):

src/alfasim_sdk/_internal/cli.py
```diff
@@ -104,6 +109,11 @@ def new(dst, caption, plugin_id, author_name, author_email):
         "}",
     ]
 
+    requirements = {
+        "sdk_version": get_required_sdk_version(),
+        "python_version": get_required_python_version(),
```
cauebs, line 114: Have we already committed to these keys? I believe "version" is implied and think we should use `"python"` and `"alfasim_sdk"`.

PR #473, opened 2025-08-25
Drop support for explicit Restart Files and add support for remote Restart Points

cauebs:


tests/alfacase/test_case_description.py
```diff
@@ -379,6 +381,56 @@ def test_check_restart_file_exists(default_case, tmp_path):
...
+
+    # Or with the other.
+    attr.evolve(
+        case,
+        physics=case_description.PhysicsDescription(
+            initial_condition_strategy=constants.InitialConditionStrategyType.Restart,
+            restart_filepath=restart_filepath,
+            restart_point_key=None,
+        ),
+    ).ensure_valid_references()
```
nicoddemus, line 431: Check the case where the file is set but does not exist.

nicoddemus (approved):

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -3427,6 +3429,23 @@ class TracersDescription:
     )
 
 
+class RestartPointLocation(Enum):
+    Local = "local"
+    Remote = "remote"
+
+
+@attr.s(frozen=True, kw_only=True)
```
BeneBr, line 3437: Why not ```dataclass```

nicoddemus: I think most of the sdk uses `attrs`, so likely he followed suit.

PR #482, opened 2025-09-25
Add steady state guess initialization configuration option

caio-ramos:
This PR introduces a new configuration option for controlling the steady state initial guess calculation strategy in ALFAsim simulations.

**Changes Made**:
- **Added new enum `SteadyStateGuessInitializationType`** with two options:
  - `QuasiHydrostatic`: Uses quasi-hydrostatic initial guess calculation (default)
  - `UserDefined`: Uses user-defined initial guess calculation
- **Enhanced `PhysicsDescription`** class with new field:
  - `steady_state_guess_initialization`: Controls the steady state guess initialization strategy
- **Updated documentation and schemas** to reflect the new configuration option

ASIM-6350

### DIFF
--- a/src/alfasim_sdk/_internal/constants.py
+++ b/src/alfasim_sdk/_internal/constants.py
@@ -569,7 +569,7 @@ class SteadyStateGuessInitializationType(Enum):
     """
     Options for steady state initial guess calculation strategy.
     """
-    
+
     # QuasiHydrostatic: Uses quasi-hydrostatic initial guess calculation
     QuasiHydrostatic = "quasi_hydrostatic"
     # UserDefined: Uses user-defined initial guess calculation
--- a/src/alfasim_sdk/__init__.py
+++ b/src/alfasim_sdk/__init__.py
@@ -216,6 +216,7 @@ from alfasim_sdk._internal.constants import (
     SimulationModeType,
     SimulationRegimeType,
     SolidsModelType,
+    SteadyStateGuessInitializationType,
     SurfaceTensionType,
     SurgeVolumeTimeMode,
     TableInputType,
@@ -224,7 +225,6 @@ from alfasim_sdk._internal.constants import (
     ValveType,
     WellConnectionPort,
     WellIndexPhaseType,
-    SteadyStateGuessInitializationType,
 )
 
 # Plugins: Context


src/alfasim_sdk/_internal/constants.py
```diff
@@ -563,3 +563,15 @@ class FluidMaterialConvectionCorrelation(Enum):
...
+class SteadyStateGuessInitializationType(Enum):
+    """
+    Options for steady state initial guess calculation strategy.
+
+    - QuasiHydrostatic: Uses quasi-hydrostatic initial guess calculation
+    - UserDefined: Uses user-defined initial guess calculation
+    """
+
+    QuasiHydrostatic = "quasi_hydrostatic"
+    UserDefined = "user_defined"
```
ggrbill, line 577: I think we use the term `From table` in the transient solver.
Please, check with Vinicius or Ana which name they prefer.

src/alfasim_sdk/_internal/constants.py
```diff
@@ -563,3 +563,15 @@ class FluidMaterialConvectionCorrelation(Enum):
     HasanKabir2002 = "Hasan and Kabir 2002"
     FuzongZhou2013 = "Fuzong Zhou 2013"
     Holman1981 = "Holman 1981"
+
+
+class SteadyStateGuessInitializationType(Enum):
```
ggrbill, line 568: I would suggest `SteadyStateInitialGuessType`. But feel free to disagree.

ggrbill (approved):

cauebs (approved):

src/alfasim_sdk/_internal/constants.py
caio-ramos: In `src/alfasim_sdk/_internal/constants.py` the enum for the initial condition strategy for the transient is available:
```python
class InitialConditionStrategyType(Enum):
    Constant = "constant"
    SteadyState = "steady_state"
    Restart = "restart"
```
At the `sdk` level, `"constant"` means the `"from table"` label at the GUI level.
So I think we could define our label here and then decide how it will appear at the GUI level.

src/alfasim_sdk/_internal/constants.py
caio-ramos: In `src/alfasim_sdk/_internal/constants.py` the enum for the initial condition strategy for the transient is available:
```python
class InitialConditionStrategyType(Enum):
    Constant = "constant"
    SteadyState = "steady_state"
    Restart = "restart"
```
At the `sdk` level, `"constant"` means the `"From table"` label at the GUI level.
So I think we could define our label here and then decide how it will appear at the GUI level.

src/alfasim_sdk/_internal/constants.py
```diff
@@ -563,3 +563,15 @@ class FluidMaterialConvectionCorrelation(Enum):
...
+    """
+    Options for steady state initial guess calculation strategy.
+    
+    - QuasiHydrostatic: Uses quasi-hydrostatic initial guess calculation
+    - UserDefined: Uses user-defined initial guess calculation
```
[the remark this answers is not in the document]
caio-ramos: Thanks, done!

PR #485, opened 2025-10-01
Add Friction Factor Correlation Model Support to ALFACase

caio-ramos:
Implements support for friction factor correlation models in ALFACase format, enabling users to specify custom friction factor correlations through plugins or use ALFAsim's default correlation.

Key Changes
----
**New Enum Added:**

- `FrictionFactorCorrelationModelType` with NoModel and FromPlugin options

**PhysicsDescription Extended:**

- `friction_factor_correlation_model`: Select correlation model type
- `friction_factor_correlation_plugin_id`: Specify plugin ID for custom correlations

**Technical Updates:**

- Updated ALFACase schema validation
- Added public API exports
- Updated documentation and tests

ASIM-5877


src/alfasim_sdk/_internal/constants.py
```diff
@@ -169,6 +169,21 @@ class SolidsModelType(Enum):
...
+    Informs which friction factor correlation model should be used:
+
+    - ModelDefault:
+        From ALFAsim's internal default friction factor correlation.
+
+    - FromPlugin:
+        Friction factor correlation is calculated from plugin hooks with an external implementation.
+    """
+
+    ModelDefault = "from_package"
```
nicoddemus, line 183: This is a strange name, perhaps `internal` would make more sense?


```suggestion
    Internal = "internal"
```

src/alfasim_sdk/_internal/constants.py
```diff
@@ -169,6 +169,21 @@ class SolidsModelType(Enum):
     FromPlugin = "from_plugin"
 
 
+class FrictionFactorCorrelationModelType(Enum):
+    """
+    Informs which friction factor correlation model should be used:
+
+    - ModelDefault:
+        From the friction factor correlation associated with the CorrelationPackageType.
+
```
caio-ramos, line 178: @nicoddemus, instead of changing to `internal`, what do you think of this explanation in the docstring?

BeneBr (approved):

src/alfasim_sdk/_internal/constants.py
BeneBr: His suggestion is about of that the "Key" ```ModelDefault ``` has a different value ```from_package```. I am with him, because these things need to have the same "context".

caio-ramos: Thanks for the explanation, now I understand the problem. Just like it is done for other enums, I defined
```
ModelDefault: "model_default"
```

PR #559, opened 2026-07-31
[ASIM-6603] Expose wall_layer_n_internal_nodes and formation refinement via alfacase

eniohayashi:
ASIM-6603

## Summary

Exposes two previously hardcoded/env-var-driven refinements as alfacase-only inputs (no GUI):

- `wall_layer_n_internal_nodes` on `EnvironmentDescription` (shared by pipes and wells): refines the radial mesh within each wall layer for the transient wall heat transfer models.
- `formation_n_layers` / `formation_thickness_ratio`, also on `EnvironmentDescription`: control the discretization of the rock formation surrounding a well for heat transfer purposes. These were briefly added to `FormationDescription` and then moved to `EnvironmentDescription` to match where `wall_layer_n_internal_nodes` lives.


ArleyAllesCruz (approved):
Nice work!

src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -1999,6 +1999,11 @@ class EnvironmentDescription:
     tvd_properties_table: list[EnvironmentPropertyDescription] = attrib_instance_list(
         EnvironmentPropertyDescription
     )
+    wall_layer_n_internal_nodes: int = attr.ib(default=0, validator=instance_of(int))
+    formation_n_layers: int = attr.ib(default=8, validator=instance_of(int))
```
BeneBr: This values must be available to be used in parametric simulations right?  So it is necessary to use the ```ArrayDescriptionValue``` or even better create a new type like:
```python
type IntExpression = int | str
```
And use this type here. Also, it is necessary to update the ```study_from_case``` and ```case_from_study``` in ```alfasim_core```.

I believe that after you update it, the ```type-check```will be failed in ```alfasim_core```, so if you preffer, create a task and assign it to me, so i will update it.

src/alfasim_sdk/_internal/alfacase/schema.py
```diff
@@ -1647,6 +1647,12 @@
         ),
         Optional("md_properties_table"): Seq(environment_property_description_schema),
         Optional("tvd_properties_table"): Seq(environment_property_description_schema),
+        Optional("wall_layer_n_internal_nodes"): Int(),
+        Optional("formation_n_layers"): Int(),
```
BeneBr: It will be updated after change the type as i mentioned before.

src/alfasim_sdk/_internal/alfacase/case_description.py
eniohayashi: Done. Also made the required changes in alfasim_core.

src/alfasim_sdk/_internal/alfacase/schema.py
eniohayashi: Done.

src/alfasim_sdk/_internal/alfacase/case_description_attributes.py
```diff
@@ -158,6 +171,7 @@ def eval_expressions(self, namespace: Mapping[str, float]) -> Curve:
 ArrayLike: TypeAlias = tuple[Sequence[Number], str] | Array
 CurveLike: TypeAlias = tuple[ArrayLike, ArrayLike] | Curve | CurveExpression
 FloatDescriptionType: TypeAlias = float | FloatExpression
+IntDescriptionType: TypeAlias = int | IntExpression
 ArrayDescriptionType: TypeAlias = Array | ArrayExpression
```
BeneBr, line 175: Dont need to change, but it can be used as is now in python 3.12
```suggestion
type ArrayLike = tuple[Sequence[Number], str] | Array
type CurveLike = tuple[ArrayLike, ArrayLike] | Curve | CurveExpression
type FloatDescriptionType = float | FloatExpression
type IntDescriptionType = int | IntExpression
type ArrayDescriptionType = Array | ArrayExpression
```

src/alfasim_sdk/_internal/alfacase/case_to_alfacase.py
```diff
@@ -146,6 +148,9 @@ def ObtainDictKey(item: Array | ArrayExpression) -> str:
     if isinstance(value, FloatExpression):
         return value.expr
 
+    if isinstance(value, IntExpression):
+        return value.expr
+
```
BeneBr: A match case statement will be nice here. But it is up to you.

BeneBr (approved):

src/alfasim_sdk/_internal/alfacase/case_to_alfacase.py
eniohayashi: I merged both checks into a single if.

PR #560, opened 2026-08-04
[ASIM-6798] Addition of porosity and por fluid material

ArleyAllesCruz:
### Motivation
The goal was to add the porosity information to the formation layer to improve its properties. The modified properties were:
* Density;
* Thermal conductivity;
* Heat capacity;
---
 ### Additional information
* The `porosity` and `pore_fluid_material` are hardcoded to 0.0 and "Brine" in a first moment. In the future such variables must be added in the GUI;


src/alfasim_sdk/_internal/alfacase/case_description.py
```diff
@@ -2282,6 +2282,12 @@ class FormationLayerDescription:
     name: str = attr.ib(validator=instance_of(str))
     start: ScalarDescriptionType = attrib_scalar(category="length")
     material: str | None = attr.ib(default=None, validator=optional(instance_of(str)))
+    porosity: ScalarDescriptionType = attrib_scalar(
+        category="dimensionless", default=Scalar("dimensionless", 0.0, "-")
+    )
+    pore_fluid_material: str = attr.ib(
+        default="Brine", validator=optional(instance_of(str))
+    )
```
BeneBr: It will have some more options? If yes, i believe that the correct is to create a ```Enum``` class for it:
```python
class PoreFluidMaterial(Enum):
     Brine = 'Brine'
     ...........
```
So when you use it in your code you can use a ```match case``` statement like this:
```python
match pore_fluid_material:
      case PoreFluidMaterial.Brine:
             # do something
      case PoreFluidMaterial.AnotherOptions
             # do another thing
      case unreachable
            assert_never(unrechable)
```
The unrechable statement will be caugh by ```Mypy``` when a modification in PoreFluidMaterial occur.

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -635,10 +635,18 @@
     target="pressure_node",
 )
 FORMATION_LAYER_DESCRIPTION = case_description.FormationLayerDescription(
-    name="f", start=Scalar(1, "m"), material="Carbon Steel"
+    name="f",
+    start=Scalar(1, "m"),
```
BeneBr: Please define the category for it because exists other categories that use ```m``` as unit like ```diameter```.

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -635,10 +635,18 @@
...
+    material="Carbon Steel",
+    porosity=Scalar(0.0, "dimensionless"),
```
BeneBr:
```suggestion
    porosity=Scalar(value=0.0, category"dimensionless", unit="-"),
```

tests/common_testing/alfasim_sdk_common_testing/filled_case_descriptions.py
```diff
@@ -635,10 +635,18 @@
...
+    pore_fluid_material="Brine",
 )
 FORMATION_LAYER_DESCRIPTION_1 = case_description.FormationLayerDescription(
-    name="f (1)", start=Scalar(1, "m"), material="Carbon Steel"
+    name="f (1)",
+    start=Scalar(1, "m"),
```
BeneBr: Same as above.

src/alfasim_sdk/_internal/alfacase/case_description.py
BeneBr: Also, it must be updated in ```configure_study_from_case``` and ```configure_case_from_study``` in core module.

BeneBr (changes requested):

BeneBr (approved):

BeneBr:
I believe that you need to rebase with master to update the pip ```requirements.txt``` to get the new pinned version of ```cog``` lib.

nicoddemus (approved):

PR #566, opened 2026-09-02
[ASIM-6850] Add dedicated roughness unit category

BeneBr:
Pipe/wall roughness (absolute roughness) was registered under the generic length category, sharing its full unit list even though roughness magnitudes are typically um-mm, much smaller than other length inputs.

Register a roughness category reusing length's dimension/conversions, with a curated unit list (m, mm, cm, um, in). Retag the case description schema and alfacase loaders for the CasingSection/Tubing/ OpenHole/Wall/PipeSegments roughness fields from length to roughness, and update fixtures/regression snapshots accordingly.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,6 +5,8 @@ CHANGELOG
 1.10.0 (UNRELEASED)
 ==================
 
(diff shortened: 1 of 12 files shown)


nicoddemus (approved):
CHANGELOG?

BeneBr:
> CHANGELOG?

Updated.

nicoddemus (approved):

PR #569, opened 2026-09-21
[ASIM-6883] Add a strategy to control timestep

ArleyAllesCruz:
## Motivation

Currently, the adaptive timestep for transient simulations is governed solely by the CFL condition. While this works well for the majority of scenarios, it leads to non-physical behavior in wall heat simulations under stagnant conditions (e.g., during a shut-in stage). Specifically, weird temperature peaks are observed in the profiles, making it necessary to introduce an additional numerical stability limit: the Fourier condition.

## What changed

* Added a Fourier-condition timestep calculation (_get_wall_Δt in time_stepper.py) for pipes/annuli when transient_bidimensional wall model is being used. The Fourier Δt only takes over when the pipe's average flow velocity is below a small threshold (near-stagnant flow). Also, the whole strategy is gated behind a new numerical option, `enable_fourier_timestep`, defaulting to **False**.
* Added 4 new unit tests in `test_time_stepper.py`, covering: 
  * Fourier limiting the timestep under stagnant flow (shut-in scenario);
  * CFL remaining in control while flowing; 
  * The option being ignored for non-`transient_bidimensional` wall models;
  * The option being ignored when disabled (the default).

## Additional information

* It will be necessary to add such option in the GUI.
* Manually validated against SPH-8 case. 

## Related task
ASIM-6883


BeneBr (changes requested):
Please update the changelog.

CHANGELOG.rst
```diff
@@ -6,6 +6,7 @@ CHANGELOG
 ==================
 
 * Add ``roughness`` category to the unit system, with a curated set of units (``m``, ``mm``, ``cm``, ``um`` and ``in``) more appropriate for roughness magnitudes than the full ``length`` unit list. ``inner_roughness`` (in ``CasingSectionDescription``, ``TubingDescription``, ``OpenHoleDescription`` and ``WallDescription``) and ``roughnesses`` (in ``PipeSegmentsDescription``) now use this category instead of ``length``.
+* Add ``enable_fourier_timestep`` attribute to ``NumericalOptionsDescription``. It controls whether pipes using the transient bidimensional wall model may have their timestep governed by the Fourier condition instead of only CFL. Such condition is apply to conditions which fluid is (near) stagnant (e.g. during a well shut-in).
```
eniohayashi, line 9:
```suggestion
* Add ``enable_fourier_timestep`` attribute to ``NumericalOptionsDescription``. It controls whether pipes using the transient bidimensional wall model may have their timestep governed by the Fourier condition instead of only CFL. The Fourier timestep control is applied when the fluid is (nearly) stagnant, such as during a well shut-in.
```

eniohayashi (approved):

BeneBr (approved):

